PR #13, opened 2016-07-29
adding garbagetruck

njharter:


garbagetruck/garbagetruck.go
```diff
@@ -0,0 +1,59 @@
+package garbagetruck
+
+import (
+	"github.com/ian-kent/go-log/logger"
```
kraj: this is MIT licensed external component. Where do we represent this information when we build webpa ?

schmidtw: Good catch - we need to collect all the licenses & ensure we meet them.
Will come up with a solution and circle back.

PR #26, opened 2016-08-26
Feature/support kid

johnabass:
Apologies for how massive this is:

The top-level of code is in the handler package, specifically AuthorizationHandler.  That is an Alice-style decorator.  In addition to tests, I supplied an example.  Let me know if you think more examples would be better.

The middle-layer of code is the secure package.  That handles parsing tokens, both Basic and Bearer (which carry JWTs).  The secure package also provides integration with SermoDigital, which is a library for working with JWTs.

The bottom layer are the secure/key and resource packages.  Those handle loading keys for validation.  Most often, this will be a public key loaded from a SATS endpoint.

The resource package is heavily inspired by Spring's resource API.   It's not a full port.  Just the parts I thought would work well as idiomatic golang.


kdamara:
Great documentation and tests!

PR #33, opened 2016-09-29
Feature/wrp refactor

johnabass:
Massive overhaul of this package:

(1) There's just (1) Message type.
(2) There are interfaces that front ugorji to ease testing and keep external code free of transitive dependencies.
(3) Reduce the number of objects required for each serialization.
(4) Expose the ugorji methods, such as Reset(), that enable object reuse.


wrp/message.go
```diff
@@ -0,0 +1,114 @@
...
+// Valid performs a basic validation check on a given message
+func (m *Message) Valid() error {
+	switch m.Type {
+	case AuthMessageType:
+		if m.Status == nil {
+			return fmt.Errorf("Missing status for message type: %s", m.Type)
+		}
+
+	case SimpleRequestResponseMessageType:
+		fallthrough
```
schmidtw: This should validate that `transaction_uuid` should be non-`nil`

johnabass: Ok, I'll make that change.  I wasn't sure if we required a transaction UUID.

PR #43, opened 2017-01-10
Feature/claims validator

njharter:


handler/authorizationHandler.go
```diff
@@ -74,8 +75,12 @@ func (a AuthorizationHandler) Decorate(delegate http.Handler) http.Handler {
 			WriteJsonError(response, forbiddenStatusCode, message)
 			return
 		}
-
-		valid, err := a.Validator.Validate(token)
+		
+		ctx := *new(context.Context)
```
johnabass: don't create a Context like this.  Per the API docs, use `context.Background()`

https://godoc.org/context#Background

secure/validator.go
```diff
@@ -105,7 +108,31 @@ func (v JWSValidator) Validate(token *Token) (valid bool, err error) {
...
+			
+			if len(pieces) == 5     &&
+			   pieces[0] == "x1"    && 
+			   pieces[1] == "webpa" && 
+			  (pieces[4] == "all" || pieces[4] == ctx.Value("method")) {
+				
+				claimPath := fmt.Sprintf("/%s/[^/]+/%s", pieces[2],pieces[3])
+				match_capPath, _ := regexp.MatchString(claimPath, ctx.Value("path").(string))
+				if match_capPath {
+					valid_capabilities = true
```
johnabass: Since you return below when this is true, why not just immediately return here?  Can eliminate a variable and a little logic.

secure/validator.go
```diff
@@ -105,7 +108,31 @@ func (v JWSValidator) Validate(token *Token) (valid bool, err error) {
...
+		for _, cap := range caps {
+			pieces := strings.Split(cap, ":")
+			
+			if len(pieces) == 5     &&
+			   pieces[0] == "x1"    && 
+			   pieces[1] == "webpa" && 
+			  (pieces[4] == "all" || pieces[4] == ctx.Value("method")) {
+				
+				claimPath := fmt.Sprintf("/%s/[^/]+/%s", pieces[2],pieces[3])
+				match_capPath, _ := regexp.MatchString(claimPath, ctx.Value("path").(string))
```
johnabass: Instead of regular expressions, why not use file globbing?  It's simpler and much cheaper to execute.

https://godoc.org/path/filepath#Match

schmidtw: When I wrote the claims on the board I used regex.  I'll change the claims to be globs.

schmidtw: Looked into this more & we're going to stick with regex.

PR #69, opened 2017-03-24
Add a skeleton for the webhooklistener code.

schmidtw:


webhooklisteners/webhooksharedlist.go
```diff
@@ -0,0 +1,35 @@
+package webhooklisteners
+
+import (
+//"time"
+)
+
+// WebHookSharedListFactory provides the input values to create a
+// WebHookSharedList service object with New()
+type WebHookSharedListFactory struct {
```
johnabass: Why have a factory?  It seems like all you need is a simple list type.

webhooklisteners/webhooksharedlist.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Add adds a WebHookListener to the shared slist
+func (w *WebHookSharedList) Add(item WebHookListener) (err error) {
+	// TODO the work of sending this to AWS
+	return
+}
+
+// Listen adds a listener function to be notifed when there is a change in the
+// list.  The entire list is given to the listener for processing.
+func (w *WebHookSharedList) Listen(func([]WebHookListener)) (err error) {
```
johnabass: Is this intended to be some sort of goroutine?  If so, I wouldn't recommend it.  A slice guarded by a sync.RWMutex is faster, better, and easier to maintain.

webhooklisteners/webhooksharedlist.go
```diff
@@ -0,0 +1,35 @@
...
+// New creates a new WebHookSharedList service object
+func (w *WebHookSharedListFactory) New() (out *WebHookSharedList, err error) {
+	// TODO do the work
+	return
+}
+
+// WebHookSharedList is the interface that provides the shared list of webhooks
+// across the system
+type WebHookSharedList struct {
+	// TODO add parameters needed to operate the service
```
johnabass: If all you need is a simple slice, consider just having this as `type WebHookSharedList []WebHookListener`

schmidtw:
Do not want.

PR #77, opened 2017-04-09
Pruned this code

johnabass:
Can we delete this package?


schmidtw:
I think so - let me double check with Nick.

njharter:
Let me double check.  I think it is still in use.

On Apr 10, 2017 9:33 AM, "Weston Schmidt" <notifications@github.com<mailto:notifications@github.com>> wrote:

I think so - let me double check with Nick.

njharter:
it is still in use with the current talaria.  as long as there is no need to use garbagegtruck in the future and we don't have to bump the webpa-common version for talaria we would be okay to remove it.  we could play it safe and remove it after we deploy new talaria. :)

johnabass:
Routinely calling the garbage collector explicitly is never necessary.  The real fix is to fix whatever issues the code has that cause memory issues.  I understand that we were just trying anything to make talaria work, but that should be the exception not the rule.

Going forward with talaria 2.0, this should never be used.  If there are issues with talaria 2.0, and it's new software so there probably will be, we should fix them instead of using this package.

njharter:
agreed

PR #84, opened 2017-04-21
Feature webhook sns

shilpa24balaji:
Added SNS Handlers
Added AWS to glide
Integrated AWS package with webhook
Added lock to synchronize access to SNS SubscriptionArn

### DIFF
--- a/webhook/aws/json_helper.go
+++ b/webhook/aws/json_helper.go
@@ -1,0 +1,29 @@
+package aws
+
+import (
+	"encoding/json"
+	"errors"
+	"net/http"
+	"io/ioutil"
+)
+
+
+var (
+	ErrJsonEmpty   = errors.New("JSON payload is empty")
+)
+
+func DecodeJSONMessage(req *http.Request, v interface{}) ([]byte, error) {
+	
+	payload, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		return nil, err
+	}
+	if len(payload) == 0 {
+		return nil, ErrJsonEmpty
+	}
+	err = json.Unmarshal([]byte(payload), v)
+	if err != nil {
+		return nil, err
+	}
+	return payload, nil
+}
--- a/webhook/aws/sns.go
+++ b/webhook/aws/sns.go
@@ -1,6 +1,20 @@
 package aws
 
-import ()
+import (
+	"github.com/Comcast/webpa-common/logging"
+	"github.com/gorilla/mux"
+	"github.com/spf13/viper"
+	"strings"
+	"net/http"
+	"net/url"
+	"sync/atomic"
+	
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/service/sns"
+	"github.com/aws/aws-sdk-go/service/sns/snsiface"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
 
 type AWSConfig struct {
 	AccessKey string    `json:"accessKey"`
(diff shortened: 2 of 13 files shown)


webhook/aws/sns.go
```diff
@@ -15,3 +28,120 @@ type SNSConfig struct {
...
+func (ss *SNSServer) PrepareAndStart() {
+	ss.Subscribe()
+}
+
+// Returns true if the SNS is ready to accept notifications
+// Synchronized using read lock
+func (ss *SNSServer) IsReady() bool {
+	var ready bool
+	ss.RLock()
+	if !strings.EqualFold("pending confirmation", "") && !strings.EqualFold("pending confirmation", ss.SubscriptionArn) {
```
schmidtw: Why perform a check against "pending confirmation" and "" ?

webhook/aws/json_helper.go
```diff
@@ -0,0 +1,57 @@
...
+
+	payload, err := ioutil.ReadAll(req.Body)
+	req.Body.Close()
+	if err != nil {
+		return nil, err
+	}
+	if len(payload) == 0 {
+		return nil, ErrJsonEmpty
+	}
+	err = json.Unmarshal([]byte(payload), v)
```
schmidtw: As written, any numbers will be returned as float64 vs a number.  You may want to use func UseNumber() if you need an accurate integer.

webhook/aws/sns.go
shilpa24balaji: SubscriptionArn will be "" or "pending confirmation" while AWS has not yet sent Subscription Confirmation Message or we have not yet saved it. Once we subscribe to SNS topic we are supposed to receive this. During this time if publish is called it will not work as SNS is not yet ready to handle notifications. So added this IsReady() function and SubscriptionArn value check.

webhook/aws/json_helper.go
johnabass: Generally speaking, you're exactly right.

However, in this particular case, `DecodeJsonPayload` is being called with a `*SNSMessage` argument for the `v` parameter.  That means the `encoding/json` package will emit numbers as whatever the struct member types are.

webhook/aws/sns.go
```diff
@@ -15,3 +28,120 @@ type SNSConfig struct {
...
+// and ready to receive AWS SNS POST messages
+// subscribe to the SNS topic, wait for snsReady
+// validate the confirmation SubscriptionArn
+func (ss *SNSServer) PrepareAndStart() {
+	ss.Subscribe()
+}
+
+// Returns true if the SNS is ready to accept notifications
+// Synchronized using read lock
+func (ss *SNSServer) IsReady() bool {
```
johnabass: This functionality is really better implemented as a `sync.WaitGroup`, an `atomic.Value`, or a channel.  Those approaches allow other goroutines to wait until things are ready without spinning on a boolean variable.

webhook/aws/sns_handler.go
```diff
@@ -0,0 +1,243 @@
...
+func (ss *SNSServer) SetSNSRoutes(urlPath string, r *mux.Router, handler http.Handler) {
+	
+	r.HandleFunc(urlPath, ss.SubscribeConfirmHandle).Methods("POST").Headers("x-amz-sns-message-type", "SubscriptionConfirmation")
+	r.Handle(urlPath, handler).Methods("POST").Headers("x-amz-sns-message-type", "Notification")
+	r.HandleFunc(urlPath, ss.UnsubscribeConfirmHandle).Methods("POST").Headers("x-amz-sns-message-type", "UnsubscribeConfirmation")
+
+}
+
+// Subscribe to AWS SNS Topic to receive notifications
+func (ss *SNSServer) Subscribe() bool {
```
johnabass: If other code needs to wait until the subscription is ready, you can use better golang-isms for this.  For example, this function could return a channel or a `sync.WaitGroup` that are used to signal that the subscription is ready.

webhook/aws/sns_handler.go
```diff
@@ -0,0 +1,243 @@
...
+		TopicArn: aws.String(msg.TopicArn), // Required
+	}
+	resp, err := ss.SVC.ConfirmSubscription(params)
+	if err != nil {
+		log.Error("SNS confirm error %v", err)
+		// TODO return error response
+		return
+	}
+
+	log.Debug("SNS confirm response: %v", resp)
```
johnabass: I would also recommend logging things to be done in Decorators.  Log statements make the code noisy, but of course you still want to log.  Separating the logging into a Decorator makes this handler's code easier to read and lets you change what's logged without affecting your core business logic.

webhook/aws/sns_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	ResponseJson(`{"message":"ok"}`, rw)
+
+}
+
+// Decodes SNS Notification message and returns 
+// the actual message which is json webhook content
+func (ss *SNSServer) NotificationHandle(rw http.ResponseWriter, req *http.Request) []byte {
+	log := ss.logger()
+	defer func() {
+		if r := recover(); r != nil {
```
johnabass: As a general rule, you don't need to do `recover` blocks in `http.Handler` code.  That's because the golang `http.Server` implementation traps panics and returns 500 responses.  The only time you'd need to recover here is if you wanted custom behavior, such as a different HTTP response code.

webhook/aws/sns_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	log := ss.logger()
+	log.Debug("SNS PublishMessage called %v ", message)
+	defer func() {
+		if r := recover(); r != nil {
+			log.Error( "SNS PublishMessage recover error %v", r )
+		}
+	}()
+
+	// check if ready to accept messages before publishing message.
+	if !ss.IsReady() {
```
johnabass: If you implement things with channels, then instead of checking readiness here you would just post your message to a channel.  A separate goroutine would consume the channel and publish to AWS if and only if the subscription was ready.  This will perform at least an order of magnitude better than locking and checking a variable.

webhook/aws/sns_handler.go
```diff
@@ -0,0 +1,243 @@
...
+			MSG_ATTR: { // Required
+				DataType:    aws.String("String"), // Required
+				StringValue: aws.String(ss.Config.Env),
+			},
+		},
+		MessageStructure: aws.String("messageStructure"),
+		Subject:          aws.String("subject new"),
+		TopicArn:         aws.String(ss.Config.Sns.TopicArn),
+	}
+	resp, err := ss.SVC.Publish(params)
```
johnabass: .... continuing from above:  here you would do `somechannel <- params` and then you're done.  You would then need a goroutine running that looked roughly like:

```
func publish(svc *sns.SNS, input <-chan *sns.PublishInput, output chan<- *sns.PublishOutput) {
    // if you want this for loop to exit, just close(input) from another goroutine
    for publishInput := range input {
        response, err := svc.Publish(publishInput)
        // logging as desired

        if err == nil {
            // this would, of course, require another goroutine to consume the output
            // I don't know if you need this, this is just an example
            // you could just log the response and drop it
            output <- response
        }
    }
}
```

webhook/aws/sns.go
johnabass: shouldn't this if statement be:
`if !strings.EqualFold("", ss.SubscriptionArn) && !strings.EqualFold("pending confirmation", ss.SubscriptionArn) ...`

shilpa24balaji: Oh yes you are right, somehow it never occurred to that I am comparing 2 const strings

PR #91, opened 2017-05-10
adding StartConfig into Factory.  startconfig default test values moved into code. added Publish method to Registry type for sns message publishing.

njharter:


johnabass:
This looks good!

PR #92, opened 2017-05-12
Webhook factory notifier tests & SNS Validation integration

shilpa24balaji:
End to end tests, using factory to Notifier Subscribe, SubscriptionConfirmation, Publish, receive notification and update webhook list

### DIFF
--- a/webhook/aws/sns.go
+++ b/webhook/aws/sns.go
@@ -36,6 +36,7 @@ type SNSServer struct {
 	subscriptionData 	chan string
 	SVC             	snsiface.SNSAPI
 	SelfUrl         	*url.URL
+	SNSValidator
 	logging.Logger
 	notificationData	chan string
 }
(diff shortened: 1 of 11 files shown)


webhook/aws/sns.go
```diff
@@ -81,7 +82,10 @@ func NewSNSServer(v *viper.Viper) (ss *SNSServer, err error) {
 
 // NewNotifier creates Notifier instance using the viper config
 func NewNotifier(v *viper.Viper) (Notifier, error) {
-	return NewSNSServer(v)
+	if v != nil {
+		return NewSNSServer(v)
+	}
+	return nil, nil
```
johnabass: You can always return a non-nil `Notifier` instance, even if `v` is nil.

That allows clients to do `NewNotifier(nil)` to get a default instance.

webhook/aws/sns_handler_test.go
```diff
@@ -52,19 +47,17 @@ const (
...
 	
 	ss := &SNSServer{
 		Config: *awsCfg,
 		SVC: m,
 	}
-	fmt.Println(ss)
-	fmt.Println(err)
 	
 	selfURL := &url.URL{
 		Scheme:   "http",
```
johnabass: If possible, use `httptest` or similar approach to avoid hardcoding the port.  These tests will break if a client has something running on port 8090.  Best to bind to an unused port for testing.

shilpa24balaji: 8090 port was just example and not actually used in unit tests to start server. selfurl is used to subscribe to SNS as endpoint to receive notifications.

PR #103, opened 2017-06-27
Feature/webhookRegistrationFix

njharter:
corrected assert expected/actual value order.
changed test to use Registry type.
changed NOTIFY_HOOK_MSG.Message to have an array value.


shilpa24balaji:
Just a thought was it necessary to publish an array of webhook containing just 1 hook instead of just sending the hook. Webhook POST /hook is for just 1 hook, so felt publish and notification should also be 1 hook. Is there any advantage of W[1] instead of just W?

schmidtw:
You're right Shilpa - I forgot that we only need the single registration vs bulk.

PR #105, opened 2017-06-29
Header to WRP

shilpa24balaji:
HTTP header to WRP conversion

### DIFF
--- a/wrp/header_wrp.go
+++ b/wrp/header_wrp.go
@@ -1,0 +1,107 @@
+package wrp
+
+import (
+	"net/http"
+	"strings"
+	"fmt"
+	"strconv"
+)
+
+// Constant HTTP header strings representing WRP fields
+const (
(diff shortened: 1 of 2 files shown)


wrp/header_wrp.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+// Convert HTTP header to WRP generic Message
+func HeaderToWRP(header http.Header) (*Message, error) {
+	msg := new(Message)
+
+	// MessageType is mandatory
+	msgType := header.Get(MsgTypeHeader)
+	if !strings.EqualFold(msgType, "") && StringToMessageType(msgType) != MessageType(-1) {
```
schmidtw: You can cut down on the string match by assigning the output of `StringToMessageType()` to a variable before this if statement.  Then just check all at once & possibly fail.  It will slow down the failure cases, but speed up the normal case.

PR #107, opened 2017-07-06
fix for removing expired listeners.  fix for newly registered listeners with a specified duration to register correctly.

njharter:


webhook/webhook.go
```diff
@@ -126,7 +127,11 @@ func (ul *updatableList) Update(newItems []W) {
 
 		// add item
 		if !found {
-			newItem.Until = time.Now().Add(DEFAULT_EXPIRATION_DURATION)
+			if newItem.Duration > 0 && newItem.Duration < DEFAULT_EXPIRATION_DURATION {
```
schmidtw: This is not the right place to handle this.  The duration should be sanitized around line 115.  The code should look like:

`items[i].Duration` = DEFAULT_EXPIRATION_DURATION
 if newItem.Duration > 0 && newItem.Duration < DEFAULT_EXPIRATION_DURATION {
    items[i].Duration = newItem.Duration
}
items[i].Until = time.Now().Add(items[i].Duration)`

So then the code here will look like this:
`newItem.Until = time.Now().Add(newItem.Duration)`

PR #109, opened 2017-07-12
Add support for WRP to HTTP header conversion

shilpa24balaji:


wrp/header_wrp.go
```diff
@@ -9,17 +9,23 @@ import (
...
-	TrasactionUuidHeader = "X-Midt-Transaction-Uuid"
-	StatusHeader         = "X-Midt-Status"
-	RDRHeader            = "X-Midt-Request-Delivery-Reponse"
-	HeadersArrHeader     = "X-Midt-Headers"
-	IncludeSpansHeader   = "X-Midt-Include-Spans"
-	SpansHeader          = "X-Midt-Spans"
-	PathHeader           = "X-Midt-Path"
-	SourceHeader         = "X-Midt-Source"
+	MsgTypeHeader         = "X-Midt-Msg-Type"
+	ContentTypeHeader     = "X-Midt-Content-Type"
```
johnabass: This should just be the standard Content-Type header.

wrp/header_wrp.go
```diff
@@ -9,17 +9,23 @@ import (
...
+	AcceptHeader          = "X-Midt-Accept"
```
johnabass: This should just be the standard Accept header

wrp/header_wrp.go
```diff
@@ -9,17 +9,23 @@ import (
...
+	TransactionUuidHeader = "X-Midt-Transaction-Uuid"
+	StatusHeader          = "X-Midt-Status"
+	RDRHeader             = "X-Midt-Request-Delivery-Reponse"
+	HeadersArrHeader      = "X-Midt-Headers"
+	IncludeSpansHeader    = "X-Midt-Include-Spans"
+	SpansHeader           = "X-Midt-Spans"
+	PathHeader            = "X-Midt-Path"
+	SourceHeader          = "X-Midt-Source"
+
+	ErrInvalidMsgType         = "Invalid Message Type"
```
johnabass: Each of these should be constants of type `error`:

    ErrInvalidMsgType = errors.New("Invalid Message Type")

wrp/header_wrp.go
```diff
@@ -57,7 +63,7 @@ func HeaderToWRP(header http.Header) (*Message, error) {
 	if !strings.EqualFold(msgType, "") && StringToMessageType(msgType) != MessageType(-1) {
 		msg.Type = StringToMessageType(msgType)
 	} else {
-		return nil, fmt.Errorf("%s", "Invalid Message Type header string")
+		return nil, fmt.Errorf("%s", ErrInvalidMsgType)
```
johnabass: Remove the `fmt.Errorf` here in favor of `error` constants (see above)

wrp/header_wrp.go
```diff
@@ -66,16 +72,16 @@ func HeaderToWRP(header http.Header) (*Message, error) {
 	} else if msg.Type == SimpleRequestResponseMessageType || msg.Type == SimpleEventMessageType ||
 		msg.Type == CreateMessageType || msg.Type == RetrieveMessageType || msg.Type == UpdateMessageType ||
 		msg.Type == DeleteMessageType {
-		return nil, fmt.Errorf("%s", "Invalid Source header string")
+		return nil, fmt.Errorf("%s", ErrInvalidSource)
 	}
 
 	// TransactionUuid is mandatory for SimpleRequestResponse and CRUD
-	if transUuid := header.Get(TrasactionUuidHeader); !strings.EqualFold(transUuid, "") {
+	if transUuid := header.Get(TransactionUuidHeader); !strings.EqualFold(transUuid, "") {
```
johnabass: Remove all validation from this code.  This block of code should only worry about translating headers into messages.  Let the `wrp` package handle validation (which already exists).

wrp/header_wrp.go
shilpa24balaji: Ya makes sense I did as per this https://github.com/Comcast/webpa-common/issues/90

schmidtw: Not that I recall.

wrp/header_wrp.go
schmidtw: Same as above since we could go from JSON -> WRP or HTTP Headers -> WRP and then the opposite.

PR #150, opened 2017-10-03
Proof that redirect can be handled. Considering a better approach

joe94:


joe94:
Closing this PR as it is not an approach we want to take.

PR #152, opened 2017-10-05
Adding MTLS support to webpacommon

shilpa24balaji:


server/webpa.go
```diff
@@ -76,10 +81,29 @@ func (b *Basic) New(logger log.Logger, handler http.Handler) *http.Server {
 		return nil
 	}
 
+	var tlsConfig *tls.Config
+	certificateFile, keyFile := b.Certificate()
+	if len(certificateFile) > 0 && len(keyFile) > 0 && len(b.ClientCACertFile) > 0 {
```
johnabass: Theae conditions should be separate.  As written, this code requires MTLS to do HTTPS.  We should be able to configure an HTTPS server with optional MTLS.

johnabass: Oh I see ... this is only MTLS.  That's fine then.  Disregard.

schmidtw:
It looks good to me - @johnabass do you approve?

PR #153, opened 2017-10-06
Keep original path in redirect URL

joe94:


service/redirectHandler.go
```diff
@@ -46,6 +47,8 @@ func (rh *RedirectHandler) ServeHTTP(response http.ResponseWriter, request *http
 		return
 	}
 
+	instance += strings.TrimRight(request.URL.Path, "/") //keep original path with trailing '/' chars removed
```
johnabass: use `strings.TrimRight(request.RequestURI, "/")` instead.  Since `RequestURI` is what's sent in the HTTP request initial line, it will include any query string and `#` markers.

joe94: @johnabass Sounds good. Thanks for clarifying that.

PR #154, opened 2017-10-10
Add GoDoc badge

joe94:
A good way to let others know they can easily browse through the code in the different packages


joe94:
Not priority at the moment.

PR #176, opened 2017-12-01
Metrics related changes only.

joe94:


metrics/metrics.go
```diff
@@ -0,0 +1,25 @@
+package metrics
+
+import (
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/prometheus"
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
+)
+
+//Provider is intended to be a metrics API for webpa servers.
+type Provider struct {
```
johnabass: We don't want our own Provider type.  We want to use go-kit's `metrics.provider.Provider` interface and associated types.  The first version can simply use `metrics.provider.NewPrometheusProvider`.

metrics/metrics.go
```diff
@@ -0,0 +1,25 @@
+package metrics
```
johnabass: Don't call this package `metrics`, as that will force collisions with other packages we use.  Instead, prefix the name so we don't need to alias it, e.g. `webpametrics`

metrics/metrics.go
```diff
@@ -0,0 +1,25 @@
...
+
+//Provider is intended to be a metrics API for webpa servers.
+type Provider struct {
+	DefaultGathererInUse bool
+}
+
+//GetCounter returns a Counter metrics Collector
+//If MetricsTool.DefaultGathererInUse is true, it will go ahead and do the registration for such metric
+//with the prometheus defaultGatherer
+func (m *Provider) GetCounter(name, help string, labelValues []string) (counter metrics.Counter) {
```
johnabass: Again, this function isn't necessary for the first version of this package.  Just use the providers in go-kit's `metrics.provider` package.

PR #186, opened 2017-12-12
Patch/config values metrics

joe94:


johnabass:
We don't want to default the metrics stuff like this.  The global defaults are enough, and are set to values that make it obvious when there's no namespace/subsystem set.

PR #201, opened 2018-01-11
adding code for metrics in webhook

njharter:


webhook/aws/metrics.go
```diff
@@ -0,0 +1,78 @@
+package aws
+
+import (
+	"github.com/Comcast/webpa-common/xmetrics"
+	"github.com/go-kit/kit/metrics"
+)
+
+const (
+	ListSize                = "list_size_value"
```
schmidtw: prefix the labels with `webhook_`

webhook/aws/sns_handler.go
```diff
@@ -215,6 +221,7 @@ func (ss *SNSServer) NotificationHandle(rw http.ResponseWriter, req *http.Reques
 	if !ss.ValidateSubscriptionArn(subArn) {
 		// Returning HTTP 500 error such that AWS will retry and meanwhile subscriptionConfirmation will be received
 		xhttp.WriteError(rw, http.StatusInternalServerError, "SubscriptionARN does not match")
+		ss.SNSNotificationReceivedCounter(http.StatusInternalServerError)
```
joe94: Don't you need to actually increment the counter that you get like so:
`ss.SNSNotificationReceivedCounter(http.StatusInternalServerError).Inc()`

njharter: this is a function that I'm passing the value to and it updates the counter

joe94: So, you're saying this https://github.com/Comcast/webpa-common/pull/201/files#diff-ce69993c618232073431360500841444R203 call updates metric you want?

njharter: wait, you're right.  I forgot some code.  thanks.

PR #204, opened 2018-01-20
Log SAT client IDs

joe94:


secure/handler/authorizationHandler.go
```diff
@@ -7,9 +7,13 @@ import (
 
 	"github.com/Comcast/webpa-common/logging"
 	"github.com/Comcast/webpa-common/secure"
+	"github.com/SermoDigital/jose/jws"
 	"github.com/go-kit/kit/log"
 )
 
+//satClientIDKey is the key to set/get sat client IDs using contexts.
+var satClientIDKey = struct{}{}
```
johnabass: You had this right before, using a custom type:

`type contextKey struct{}`

The reason you don't want to use `struct{}{}` here as a global variable is that it can collide with other package keys.  A custom type, using `struct{}`, is canonicalized by go and is distinct from any other type.

secure/handler/authorizationHandler.go
```diff
@@ -106,8 +110,11 @@ func (a AuthorizationHandler) Decorate(delegate http.Handler) http.Handler {
 		ctx = context.WithValue(ctx, "method", request.Method)
```
johnabass: This kind of code is handled by go-kit's `transporthttp` package.  Leverage that if you can.

Also, don't use string context keys.  Use a custom type.  If you need to add lots of context variables, it's more efficient to just use a `struct` under a single key:

```
type Sat struct {
    Method string
    Path string
    ClientID string
}

// don't be afraid to pass `sat` by value.  it's efficient and ensures immutability.
func NewContextSat(ctx context.Context, sat Sat) context {
    return context.WithValue(contextKey{}, sat)
}
```

secure/handler/authorizationHandler.go
```diff
@@ -117,6 +124,7 @@ func (a AuthorizationHandler) Decorate(delegate http.Handler) http.Handler {
 			logging.MessageKey(), "request denied",
 			"validator-response", valid,
 			"validator-error", err,
+			"sat-client-id", satClientID,
```
johnabass: Don't forget that you can emit a `logger` with the `sat-client-id` prebaked:

`logger = log.With(logger, "sat-client-id", satClientID`

That way, any code that uses `logger` from there on out will get the `sat-client-id` field automatically.

PR #205, opened 2018-01-20
To add MTLS client cert validation support

shilpa24balaji:


server/viper.go
```diff
@@ -199,6 +202,10 @@ func Initialize(applicationName string, arguments []string, f *pflag.FlagSet, v
 		return
 	}
 
+	if v.IsSet(PeerVerifySuffix) {
+		webPA.Primary.PeerVerify = v.GetBool(PeerVerifySuffix)
```
johnabass: `PeerVerify` should just be part of the `WebPA` struct itself.

server/webpa.go
```diff
@@ -81,6 +83,8 @@ type Basic struct {
 	CertificateFile    string
 	KeyFile            string
 	ClientCACertFile   string
+	PeerVerify	   bool
+	PeerVerifyFunc     PeerVerifyCallback	// Callback func to add peer client cert CN, SAN validation
```
johnabass: The zero value for this variable should be valid.  Create a default verify callback that is used when this variable is `nil` and `PeerVerify` is `true`

server/webpa.go
```diff
@@ -155,6 +163,11 @@ func (b *Basic) New(logger log.Logger, handler http.Handler) *http.Server {
 				ClientAuth: tls.RequireAndVerifyClientCert,
 			}
 			tlsConfig.BuildNameToCertificate()
+
+			// Add verify peer ceritifcate callback for additional validation
+			if b.PeerVerify && b.PeerVerifyFunc != nil {
+				tlsConfig.VerifyPeerCertificate = *b.PeerVerifyFunc
```
johnabass: There's no need to dereference `PeerVerifyFunc`.  Functions are reference types in `golang`

PR #222, opened 2018-01-26
(WIP - Do not merge) provide a common prehandler

joe94:


secure/handler/provider.go
```diff
@@ -0,0 +1,89 @@
...
+	// JWTKeys is used to create the key.Resolver for JWT verification keys
+	Keys key.ResolverFactory
+
+	// Custom is an optional configuration section that defines
+	// custom rules for validation over and above the standard RFC rules.
+	Custom secure.JWTValidatorFactory
+}
+
+//GetPreHandler returns a configured tunnel with requirements for requests to pass through before they reach some main handler
+func GetPreHandler(v *viper.Viper, logger log.Logger, registry xmetrics.Registry) (preHandler *alice.Chain, err error) {
```
johnabass: Overall, this is a great idea.  Some observations:

- `GetPreHandler` isn't a good name for this.  `golang` style is to generally avoid `Get` as a method prefix, and `PreHandler` isn't clear as to it's purpose.  `InstrumentXXX` is a better name, since that's what you're doing: instrumenting another handler.
- This code shouldn't work in terms of viper.  Let client code use Viper to create the validator and whatever else is necessary.
- Returning an `alice.Chain` makes this inflexible.  A client can't very easily continue with chaining things.  A better approach is to have this function return an Alice-style constructor:

```
// This lets clients chain additional Alice-style middleware
func InstrumentHandler(logger log.Logger, registry xmetrics.Registry) func(http.Handler) http.Handler {
    // handle defaulting the logger to logging.DefaultLogger() if you like

    return func(next http.Handler) http.Handler {
        // do the decoration here
    }
}
```

- A variant of the above, if you have a lot of options, is to create an options struct and use that instead.  This works well when you have more than (2) parameters, or you think that you may have more than (2) parameters in the future:

```
type ValidatorOptions struct {
    Logger log.Logger
    Registry xmetrics.Registry
    // more stuff
}

func InstrumentHandler(o ValidatorOptions) func(http.Handler) http.Handler {
    // passing on the stack has all kinds of benefits:  a race condition on ValidatorOptions is impossible, and you can change ValidatorOptions variables as you like, e.g. if you want to default nil members

    return func(next http.Handler) http.Handler {
        // do the decoration here
    }
}
```

PR #252, opened 2018-03-23
Increase Default ReadTimeout to prevent TLS sycall timeout error

shilpa24balaji:
Default ReadTimeout of 5s is not sufficient on clients with high load average. TLS syscall errors are seen for petasos and talaria.


shilpa24balaji:
Will do this change specific for comcast hence closing

PR #265, opened 2018-04-18
fanouthttp will return the larget code from the spans.  it also does …

njharter:
…not convert codes 500and over to StatusServiceUnavailable


schmidtw:
Looks good to me.  @johnabass ?

PR #278, opened 2018-05-12
simplify test script

joe94:
This simplifies the test script by using the existing go tool capabilities to run all package tests in parallel:

According to the docs (https://golang.org/cmd/go/#hdr-Compile_packages_and_dependencies), we can configure the number of tests that can run in parallel with the `-p` flag. I left it out as the default is the available number of CPUs which seems like a good starting point. 

*Note:* "all packages" above refers to packages in `webpa-common` except those under vendor\ (https://tip.golang.org/doc/go1.9#vendor-dotdotdot)

**Dependency:** needs go version >= `1.10`. Specifically, the docs( https://golang.org/doc/go1.10#test) describe that the "`go test -coverprofile` option is now supported when running multiple tests"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go: 
-    - 1.9.2
+    - "1.10"
 
 before_install:
     - sudo pip install --user codecov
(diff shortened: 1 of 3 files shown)


joe94:
@johnabass have you been able to give this another look?

Issue #271, opened 2018-04-27
Add more information to the Connect event


Symbianx:
After looking at the logs of talaria we were able to see some really valuable information when a device connects to the server, information called the `convey` payload. Fields like `boot-time` and `hw-last-reboot-reason` are included in the log. [This is the log I am talking about](https://github.com/Comcast/webpa-common/blob/bd4c644760d8d708804f87461c513271c046de4a/device/manager.go#L158).

In terms of analytics and quality of service metrics all this information would be really useful for us. Is this something you would be willing to add?

Thanks in advance
Miguel Alexandre
NOS Inovação

schmidtw:
Good suggestion!  We added the information from the convey header to the logs to start with so that log processors can grab the information and make it available somewhat.  On the metrics front, the concern we have is the unbounded set of data that it represents.

https://prometheus.io/docs/practices/naming/#labels

It may make sense to identify/design an interface to something like redis that allows some state-full data to be gathered/reported upon outside the normal production data flow.

Symbianx:
The concern about prometheus is extremely valid.

We thought about it and we think it would be really awesome if the convey data was incorporated in the `operational` event (which already exists) and sent to all the webhook listeners, this way the listeners could take more complex and intelligent decisions based on the event.
This solution would also take the responsability of saving the data (like you said, with redis for instance) out from talaria into a service responsible for just that.

kristinaspring:
@schmidtw @Symbianx can this issue be closed?

PR #281, opened 2018-05-22
Add convey payload to connect event

Symbianx:
This change adds the convey payload to the Connect event as discussed in #271. 
When used together with Comcast/talaria#43 will allow us to take more informed actions based on the device state.

### DIFF
--- a/device/manager.go
+++ b/device/manager.go
@@ -1,6 +1,7 @@
 package device
 
 import (
+	"encoding/json"
 	"io"
 	"net/http"
 	"sync"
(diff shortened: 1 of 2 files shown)


schmidtw:
@johnabass Can we merge this or are there changes needed?

johnabass:
We had some internal discussions and had decided to delay this due to performance concerns.  Has anything changed?

Symbianx:
@johnabass @schmidtw 
Although we understand your concern about performance, we think this change would really benefit all companies using this software, being able to leverage the information contained in the convey payload, such as reboot reasons and uptime, is what makes the Connect event really valuable for extracting business indicators and providing a better costumer experience.

Can we try to workout a solution regarding the performance impact and still maintain the benefits of the pull request?
Is your concern based on the Marshalling of the object? We could move the marshalling to the line before the logging, since the logger is also marshalling to JSON, passing the already marshalled string in the log should leave us with about the same performance. What do you think?

schmidtw:
Yeah, this is a good step forward.  The biggest concern we have is actually with the other part of the PR where the internal talaria events get propagated out.  Given the volume of some of them (and the value of some & not others) we will need to add some knobs to the other PR.

PR #285, opened 2018-05-30
Fix talaria multiple registrations

Symbianx:
Hello,
In our environment we need talaria to register in consul with 2 endpoints, one for internal clients and another for the external clients. When we attempted this configuration, looking at tcpdumps we noticed that talaria registered 2 times using the same information (e.g. address, port).
This happens because, according to the spec, the reference to the `registration` may be re-used through all iterations of the loop:

> The iteration variables may be declared by the "range" clause using a form of short variable declaration (:=). In this case their types are set to the types of the respective iteration values and their scope is the block of the "for" statement; they are re-used in each iteration. If the iteration variables are declared outside the "for" statement, after execution their values will be those of the last iteration.

This PR creates a new reference to the `AgentServiceRegistration` while looping, effectively fixing the issue.

### DIFF
--- a/service/consul/environment.go
+++ b/service/consul/environment.go
@@ -100,19 +100,19 @@ func newInstancers(l log.Logger, c gokitconsul.Client, co Options) (i service.In
 func newRegistrars(l log.Logger, registrationScheme string, c gokitconsul.Client, u ttlUpdater, co Options) (r service.Registrars, closer func() error, err error) {
 	var consulRegistrar sd.Registrar
 	for _, registration := range co.registrations() {
-		localRegistration := registration
+		registration := registration
 
-		instance := service.FormatInstance(registrationScheme, localRegistration.Address, localRegistration.Port)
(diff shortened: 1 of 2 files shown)


service/consul/environment.go
```diff
@@ -100,17 +100,19 @@ func newInstancers(l log.Logger, c gokitconsul.Client, co Options) (i service.In
 func newRegistrars(l log.Logger, registrationScheme string, c gokitconsul.Client, u ttlUpdater, co Options) (r service.Registrars, closer func() error, err error) {
 	var consulRegistrar sd.Registrar
 	for _, registration := range co.registrations() {
-		instance := service.FormatInstance(registrationScheme, registration.Address, registration.Port)
```
johnabass: Idiomatic go in this case is to do `registration := registration`

Symbianx: I did not know that, thank you.

Fixed.

schmidtw:
@johnabass Can this be merged or are there other changes that you think are needed?

PR #291, opened 2018-07-13
adding wait logic for webhook/aws.  when used, this allows a service …

njharter:
…to wait for its DNS to propagate before attempting to subscribe to AWS SNS.


njharter:
this will fix
https://github.com/Comcast/webpa-common/issues/290

PR #294, opened 2018-07-23
Allow prepared messages to be enqueue asynchronously; send the auth s…

johnabass:
…tatus message through the write queue


kcajmagic:
Closing and moving forward with deprecation of `authStatus` Message

PR #302, opened 2018-08-07
fixing the possibility of `PublishMessage` hanging

kcajmagic:
- the hanging could happen by the channel not being emptied from the
  inability to talk to SNS
- two new configurations were added to viper
  - sns.channelSize the size of the channel to buffer the messages
  - sns.channelClientTimeout the timeout for adding messages to buffer

### DIFF
--- a/webhook/aws/sns_handler.go
+++ b/webhook/aws/sns_handler.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"container/list"
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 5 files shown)


webhook/aws/sns_handler.go
```diff
@@ -283,13 +284,21 @@ func (ss *SNSServer) NotificationHandle(rw http.ResponseWriter, req *http.Reques
...
 
 	// push Notification message onto notif data channel
-	ss.notificationData <- message
+	ticker := time.NewTicker(ss.channelClientTimeout)
+	defer ticker.Stop()
+
+	select {
+	case ss.notificationData <- message:
+		return nil
+	case <-ticker.C:
```
johnabass: Don't use a ticker here.  It's overkill.  Use `NewTimer` instead.

webhook/aws/sns_handler_test.go
```diff
@@ -379,6 +386,25 @@ func TestNotificationHandleError_ValidationErr(t *testing.T) {
...
+func TestPublishClientTimeout(t *testing.T) {
+	fmt.Println("\n\nTestNotificationHandleError_ValidationErr")
+
+	ss, m, _, _ := SetUpTestSNSServerWithChannelSize(t, 1)
+	ss.channelClientTimeout = 1
+
+	// should not panic
+	testPublish(t, m, ss)
+
+	defer func() {
```
johnabass: Consider using the `stretchr/testify` library, with has an `assert.Panics` assertion that will clean up this code considerabliy.

PR #316, opened 2018-09-11
Fix Content Type

githubsands:
Solves Comcast/tr1d1um#85.


githubsands:
Thanks wasn't sure what travis was barking about.

PR #317, opened 2018-09-17
fix fanout: on sequential request returning a 503

kcajmagic:
On sequential request the request would occasionally return a 503 instead of a 504. This is because the client would timeout before the context would.

### DIFF
--- a/xhttp/fanout/handler.go
+++ b/xhttp/fanout/handler.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/Comcast/webpa-common/logging"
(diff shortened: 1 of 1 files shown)


xhttp/fanout/handler.go
```diff
@@ -253,6 +254,9 @@ func (h *Handler) execute(logger log.Logger, spanner tracing.Spanner, results ch
 
 		if result.Err == context.Canceled || result.Err == context.DeadlineExceeded {
 			result.StatusCode = http.StatusGatewayTimeout
+		} else if strings.Contains(result.Err.Error(), "Client.Timeout exceeded while awaiting headers)") {
```
njharter: Where does this error come from? What happens if it changes?

PR #330, opened 2018-09-27
add ability to get first cause

kcajmagic:
Usually custom errors contain a field with an error to cascade the error up. This allows for the ability to get the root cause. For example, `*url.Error` has the field `Err` which could be a "context deadline exceeded", "context canceled" or a timeout error.


xerrors/xerrors_test.go
```diff
@@ -0,0 +1,191 @@
...
+type subErrorPtr struct {
+	err *error
+	str string
+}
+
+func (err subErrorPtr) Error() string {
+	return fmt.Sprintf("%s(%s)", err.str, (*err.err).Error())
+}
+
+func TestGetError_Int(t *testing.T) {
```
johnabass: Don't use underscores in test names.  `TestGetErrorInit` is fine as is.

PR #335, opened 2018-10-08
Adding custom resolver package

kcajmagic:
When this package is used, the transport layer will dial  a custom route instead of the original route.

# Usage:
For example, the following will dial `customA.host.com:8080` instead of `custom.host.com:9090`. 

```go
	watcher := NewConsulWatcher(&Options{
		Watch: map[string]string{"custom.host.com": "myService"},
	})

	// note: MonitorEvent is the Listener interface in the monitor package
	watcher.MonitorEvent(monitor.Event{
		Key:       service + "[tag tagA]" + "{passingOnly=true}",
		Instances: []string{"http://customA.host.com:8080"},
	})

	client := &http.Client{
		Transport: &http.Transport{
			DialContext:       xresolver.NewResolver(nil, watcher).DialContext,
			// note: DisableKeepAlives is required so when we do the request again we don't reuse the same connection.
			DisableKeepAlives: true,
		},
	}

	req, err := http.NewRequest("GET", "http://custom.host.com:9090", nil)
	assert.NoError(err)

	res, err := client.Do(req)
```

# What is added in this package:
- create `DialContext` logic for the transport layer.
- added a RoundRobin balancer
 - created a consul watcher that uses the RoundRobin balancer and updates the routes via consul.

### DIFF
--- a/xresolver/consul/consullistener.go
+++ b/xresolver/consul/consullistener.go
@@ -18,17 +18,22 @@ type Options struct {
 	// exp. { "beta.google.com" : "caduceus" }
 	Watch map[string]string `json:"watch"`
 
-	logger log.Logger `json:"-"`
+	Logger log.Logger `json:"-"`
 }
 
 type ConsulWatcher struct {
+	debugLogger log.Logger
+	infoLogger  log.Logger
+	warnLogger  log.Logger
+	errorLogger log.Logger
+
 	balancers map[string]*xresolver.RoundRobin
 	config    *Options
 }
 
 func NewConsulWatcher(o *Options) *ConsulWatcher {
-	if o.logger == nil {
-		o.logger = logging.DefaultLogger()
+	if o.Logger == nil {
+		o.Logger = logging.DefaultLogger()
 	}
 
 	balancers := make(map[string]*xresolver.RoundRobin)
@@ -38,8 +43,12 @@ func NewConsulWatcher(o *Options) *ConsulWatcher {
 		}
 	}
 	watcher := &ConsulWatcher{
-		balancers: balancers,
-		config:    o,
+		balancers:   balancers,
+		config:      o,
+		debugLogger: logging.Debug(o.Logger),
+		infoLogger:  logging.Info(o.Logger),
+		warnLogger:  logging.Warn(o.Logger),
+		errorLogger: logging.Error(o.Logger),
 	}
 
 	return watcher
@@ -57,12 +66,13 @@ func (watcher *ConsulWatcher) MonitorEvent(e monitor.Event) {
 			// find records
 			route, err := xresolver.CreateRoute(instance)
 			if err != nil {
-				logging.Warn(watcher.config.logger).Log(logging.MessageKey(), "failed to create route", logging.MessageKey(), err, "instance", instance)
+				watcher.errorLogger.Log(logging.MessageKey(), "failed to create route", logging.MessageKey(), err, "instance", instance)
 				continue
 			}
 			routes[index] = *route
 		}
 		rr.Update(routes)
+		watcher.debugLogger.Log(logging.MessageKey(), "received update event", "service", str[1], "new-routes", routes)
 	}
 }
 
@@ -70,5 +80,7 @@ func (watcher *ConsulWatcher) LookupRoutes(ctx context.Context, host string) ([]
 	if _, found := watcher.config.Watch[host]; !found {
 		return []xresolver.Route{}, errors.New(host + " is not part of the consul listener")
 	}
-	return watcher.balancers[watcher.config.Watch[host]].Get()
+	records, err := watcher.balancers[watcher.config.Watch[host]].Get()
+	watcher.debugLogger.Log(logging.MessageKey(), "looking up routes", "routes", records, logging.ErrorKey(), err)
+	return records, err
 }
--- a/xresolver/consul/consullistener.go
+++ b/xresolver/consul/consullistener.go
@@ -7,7 +7,6 @@ import (
 	"github.com/Comcast/webpa-common/service/monitor"
 	"github.com/Comcast/webpa-common/xresolver"
 	"github.com/go-kit/kit/log"
-
 	"regexp"
 )
 
@@ -27,8 +26,10 @@ type ConsulWatcher struct {
 	warnLogger  log.Logger
 	errorLogger log.Logger
 
+	config *Options
+
+	watch     map[string]string
 	balancers map[string]*xresolver.RoundRobin
-	config    *Options
 }
 
 func NewConsulWatcher(o *Options) *ConsulWatcher {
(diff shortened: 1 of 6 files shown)


xresolver/xresolver.go
```diff
@@ -0,0 +1,113 @@
...
+	"sync"
+)
+
+// Note to self: Dial is not being set for net.Resolver because that is the Dial to the DNS server.
+
+var DefaultDialer = &net.Dialer{}
+
+type resolver struct {
+	resolvers map[Lookup]struct{}
+	lock      *sync.RWMutex
```
johnabass: Same issue as above: don't use a pointer here.  Just declare it as a value.

xresolver/xresolver.go
```diff
@@ -0,0 +1,113 @@
...
+	"strconv"
+	"sync"
+)
+
+// Note to self: Dial is not being set for net.Resolver because that is the Dial to the DNS server.
+
+var DefaultDialer = &net.Dialer{}
+
+type resolver struct {
+	resolvers map[Lookup]struct{}
```
johnabass: If you are trying to do a set, use `map[Lookup]bool`.  That's the standard `golang` idiom for a set of objects.  You can then rely on zero-values: `resolvers[...]` will return `false` if no such lookup is in the map, `true` if it is.

See https://golang.org/doc/effective_go.html#maps

xresolver/xresolver_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	req, err := http.NewRequest("GET", ts.URL, nil)
+	assert.NoError(err)
+
+	res, err := client.Do(req)
+	assert.NoError(err)
+	assert.Equal(200, res.StatusCode)
+}
+
+type testCustomLookUp struct {
```
johnabass: Consider replacing this with a mock.  If `Lookup` is an interface, you can easily use `stretchr/testify` to create a mock that plays well with the `golang` `testing` package.

xresolver/consul/consullistener.go
```diff
@@ -0,0 +1,103 @@
...
+	warnLogger  log.Logger
+	errorLogger log.Logger
+
+	config *Options
+
+	watch     map[string]string
+	balancers map[string]*xresolver.RoundRobin
+}
+
+func NewConsulWatcher(o *Options) *ConsulWatcher {
```
johnabass: Don't pass `Options` as a pointer here.  This is a potential race condition, given what happens immediately after this line.  The `Logger` field is edited on a shared object.

In general, just pass `Options`-style structs by value.  I know other places have them passed by pointer, but this was due to a lack of familiarity with concurrent programming.

xresolver/types.go
```diff
@@ -0,0 +1,180 @@
...
+	robin.lock.RLock()
+	if _, found := robin.routes[route.String()]; found {
+		robin.lock.RUnlock()
+		return errors.New("addr already in rotation")
+	}
+	robin.lock.RUnlock()
+
+	// Add to our structure
+	robin.lock.Lock()
+	defer robin.lock.Unlock()
```
johnabass: In general, do the `defer ....Unlock()` first.  The reason is because `defer` has a nontrivial execution time.  When you hold a mutex, you want to minimize the length of time the lock is held.

xresolver/types.go
```diff
@@ -0,0 +1,181 @@
...
+
+func NewRoundRobinBalancer() *RoundRobin {
+	return &RoundRobin{
+		routes: make(map[string]*orderedRoute),
+	}
+}
+
+func (robin *RoundRobin) Add(route Route) error {
+	// check if exist
+	robin.lock.RLock()
```
johnabass: You should probably avoid the two `RUnlock` calls here, as this is a future bug waiting to happen.  This works better:

```
robin.lock.RLock()
_, found := robin.routes[route.String()]
robin.lock.RUnlock()
if found {
    return errors.New("addr already in rotation")
}
```

That's a lot easier to reason about.  And future bug fixes have a much lower risk of messing up the lock/unlock code.

xresolver/types.go
```diff
@@ -0,0 +1,181 @@
...
+	robin.routes[route.String()] = &orderedRoute{
+		route: route,
+		index: len(robin.routes),
+	}
+	robin.lock.Unlock()
+	return nil
+}
+
+func (robin *RoundRobin) Remove(route Route) error {
+	robin.lock.RLock()
```
johnabass: Same comment as above: don't, in general, do `RUnlock` in such a disjointed way.

xresolver/types.go
```diff
@@ -0,0 +1,181 @@
...
+	if _, found := robin.routes[route.String()]; !found {
+		robin.lock.RUnlock()
+		return errors.New("addr not found")
+	}
+	defer func() {
+		robin.lock.RUnlock()
```
johnabass: This is inherently dangerous, as it depends upon the defer side effect for the unlocking.  Better to handle this outside the defer.

The lock/unlock below, completely *within* the defer, is fine.  That's easy to reason about.

xresolver/types.go
```diff
@@ -0,0 +1,181 @@
...
+		return records, errors.New("no records available")
+	}
+
+	for _, route := range robin.routes {
+		records[route.index] = route.route
+	}
+
+	// update order
+	defer func() {
+		robin.lock.RUnlock()
```
johnabass: Same issue as above:  don't to an unlock within a defer that depends on a corresponding lock outside the defer.

xresolver/xresolver.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// Note to self: Dial is not being set for net.Resolver because that is the Dial to the DNS server.
+
+var DefaultDialer = &net.Dialer{}
+
+type resolver struct {
+	resolvers map[Lookup]bool
+	lock      sync.RWMutex
+	dialer    *net.Dialer
```
johnabass: This works better as a non-pointer member.  Below, in `NewResolver`, you copy a pointer from outside the function into this function.  The `resolver` type thus holds a reference to an object outside its scope, which can lead to odd race conditions.

If this is a non-pointer member, the struct initializer in `NewResolver` will copy it completely, thus making a separate instance.  This makes for much more bulletproof code.

xresolver/xresolver.go
```diff
@@ -0,0 +1,111 @@
...
+	"context"
+	"errors"
+	"net"
+	"strconv"
+	"sync"
+)
+
+// Note to self: Dial is not being set for net.Resolver because that is the Dial to the DNS server.
+
+var DefaultDialer = &net.Dialer{}
```
johnabass: You can eliminate this if the `resolver.dialer` field isn't a pointer.

xresolver/types.go
njharter: I would a channel be a better solution?

johnabass: no, there's no reason to force a context switch here.  a mutex is far better (or, an atomic variable).

PR #336, opened 2018-10-10
redirectHandler:  Log specified headers 

githubsands:
Logs device header and authorization header as request by Shilpa. 

Comcast/petasos#23

### DIFF
--- a/service/servicehttp/redirectHandler.go
+++ b/service/servicehttp/redirectHandler.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/Comcast/webpa-common/logging"
+	"github.com/Comcast/webpa-common/logging/logginghttp"
 	"github.com/Comcast/webpa-common/service"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


service/servicehttp/redirectHandler.go
```diff
@@ -50,6 +51,10 @@ func (rh *RedirectHandler) ServeHTTP(response http.ResponseWriter, request *http
 	instance += strings.TrimRight(request.RequestURI, "/")
 	rh.Logger.Log(level.Key(), level.DebugValue(), logging.MessageKey(), "redirecting", "instance", instance)
 
+	logHeaderFunc, ctx := logginghttp.SetLogger(rh.logger, logginghttp.Header("X-WebPA-Device-Name", "device_id"), logginghttp.Header("Authorization", "authorization"))
```
shilpa24balaji: But this will not get logged in the error case of 400 in line 39 above. I want this information during error cases only.

githubsands: okay give me a bit

PR #337, opened 2018-10-10
add failover accessor

kcajmagic:
# Overview
-  `LayeredAccessor` adds a layer of complexity on top of the `Accessor` interface.
-  `GateAccessor` combines the functionality of `Gate` and the `Accessor` interface

# Description
`LayeredAccessor` has both a primary `Accessor` and an array (under the hood its `map[string]struct{}`) of `Accessor`s. This struct will try and find an instance via the primary (aka. its own data center) then will try the fail overs.

# Changes
In order to support this new struct. Two interfaces were created `RouteTraffic` and `AccessorQueue`.

-  RouteTraffic
   -  this interface is used for determine if a an instance should be returned. For example, an accessor may return an instance but that instance may not be routable from host.
   -  The default behavior via `DefaultTrafficRouter()` is to allow all instances. 
-  AccessorQueue
   -  this interface is used to determine the order at which to try the failover instances.
   -  The default behavior via `DefaultOrder()` is _random_ in that during the iteration over the map will be the order.
-  GateAccessor
   -  uses the gate feature to determine if accessor should be used 
   -  will still return the hashed instance along with the error _"gate is closed"_

# Example Usage 
https://github.com/kcajmagic/petasos/blob/trafficFailover/src/petasos/petasos.go

### DIFF
--- a/service/accessor.go
+++ b/service/accessor.go
@@ -3,11 +3,15 @@ package service
 import (
 	"errors"
 	"fmt"
+	"math/rand"
 	"sync"
 )
 
 var (
-	errNoInstances = errors.New("There are no instances available")
+	errNoInstances     = errors.New("There are no instances available")
+	errNoFailOvers     = errors.New("no failover instances available")
+	errNoRouter        = errors.New("traffic router interface not set")
+	errFailOversFailed = errors.New("failovers could not find an instance")
 )
 
 // Accessor holds a hash of server nodes.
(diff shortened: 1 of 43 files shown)


service/accessor.go
```diff
@@ -96,3 +100,237 @@ func (ua *UpdatableAccessor) Update(a Accessor, err error) {
...
+func (r emptyQueue) Order(keys []string) []string {
+	return keys
+}
+
+type AccessorValue struct {
+	Accessor Accessor
+	Err      error
+}
+
+type LayeredAccessor struct {
```
johnabass: This should really be an interface, with an internal implementation such as `layeredAccessor`. 

Currently, this type is mutable, which means you have to have a mutex for things like the router.  By using an interface returned by a `NewLayeredAccessor` function, the returned `LayeredAccessor` can be immutable.  This, in turn, removes any chance for a deadlock now or in the future.

service/accessor.go
```diff
@@ -96,3 +100,237 @@ func (ua *UpdatableAccessor) Update(a Accessor, err error) {
...
+	accessorQueue AccessorQueue
+
+	err     error
+	primary Accessor
+
+	failover map[string]AccessorValue
+}
+
+// SetRouter will update teh router, which will determine if the accessor should return the instance or fail
+func (la *LayeredAccessor) SetRouter(router RouteTraffic) {
```
johnabass: As an example of mutability:  This method mutates the `LayeredAccessor` type, forcing you to use a mutex.

If a `NewLayeredAccessor` constructor was provided, the router can be supplied up front and thus be immutable for the life of the `LayeredAccessor` object.

service/accessorFactory.go
```diff
@@ -38,6 +41,21 @@ func NewConsistentAccessorFactory(vnodeCount int) AccessorFactory {
 	}
 }
 
+// NewConsistentAccessorFactoryWithGate produces a factory which uses consistent hashing
+// of server nodes with the gate feature.  The returned factory does not modify instances passed to it.
+// Instances are hashed as is. If the gate is closed an error saying the gate is closed will be returned
+func NewConsistentAccessorFactoryWithGate(vnodeCount int, g gate.Interface) AccessorFactory {
```
johnabass: What's the use case for this?  Why doesn't the gate work as is via Alice-style decoration?

PR #338, opened 2018-10-11
redirectHandler: get logger from context when error

githubsands:
Comcast/petasos#23


service/servicehttp/redirectHandler.go
```diff
@@ -35,6 +35,8 @@ type RedirectHandler struct {
 func (rh *RedirectHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
 	key, err := rh.KeyFunc(request)
 	if err != nil {
+		ctxLogger := logging.GetLogger(request.Context())
+		rh.Logger = ctxLogger
```
johnabass: This is a race condition.  Many goroutines will simultaneously call `ServeHTTP`.

It's also not necessary.  You can delete the `Logger` member from the `RedirectHandler`, and just use the logger from the context.

PR #339, opened 2018-10-11
Removing Port From DNS Query Question

kristinaspring:


webhook/aws/sns.go
```diff
@@ -244,7 +244,7 @@ func (ss *SNSServer) DnsReady() (e error) {
 				}
 				// otherwise, we keep trying
 				ss.metrics.DnsReadyQueryCount.Add(1.0)
-				ss.debugLog.Log(logging.MessageKey(), "checking if server's DNS is ready", "endpoint", ss.SelfUrl.Host, logging.ErrorKey(), err)
+				ss.debugLog.Log(logging.MessageKey(), "checking if server's DNS is ready", "endpoint", ss.SelfUrl.Host, logging.ErrorKey(), err, response)
```
johnabass: The `Log` method requires an even number of parameters in `key`, `value`, ... format.  The `response` at the end has no associated key.  You can just put in an ad hoc key via `"response", response`.

Also, be sure that `response` implements `fmt.Stringer`, i.e. has a `String` method.  If it doesn't, it won't output properly for all log formats.

kristinaspring: Added the "response" string per your suggestion.  I verified and response does implement `fmt.Stringer`, so that is good. Thanks!

PR #346, opened 2018-11-07
Adds device metrics for convey headers

kcajmagic:
## What changed
 - Adds the conveymetric package.
    - This package will have a new gauge for each value for a certain key.
    - Then on completion of the convey (aka. disconnect) the `MetricClosure` must be called for accurate results
 - Add conveymetric to device package to track `hw_model`

Required for [Talaria Issue](https://github.com/Comcast/talaria/issues/68)

### DIFF
--- a/convey/conveymetric/metric.go
+++ b/convey/conveymetric/metric.go
@@ -1,0 +1,67 @@
+package conveymetric
+
+import (
+	"fmt"
+	"github.com/Comcast/webpa-common/convey"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const UnknownLabel = "unknown"
+
+type MetricClosure func()
+
+type CMetric interface {
+	Update(data convey.C) (MetricClosure, error)
+	GetMetrics() []prometheus.Collector
+}
+
+func NewConveyMetric(tag string, name string) CMetric {
+	return &cMetric{
+		Tag:     tag,
+		Name:    name,
+		metrics: make(map[string]prometheus.Gauge),
+	}
+}
+
+type cMetric struct {
+	Tag     string
+	Name    string
+	metrics map[string]prometheus.Gauge
+}
+
+func (m *cMetric) Update(data convey.C) (MetricClosure, error) {
+	var gague prometheus.Gauge
+
+	key := UnknownLabel
+
+	if item, ok := data[m.Tag].(string); ok {
+		key = item
+	}
+
+	if val, found := m.metrics[key]; found {
+		gague = val
+	} else {
+		m.metrics[key] = prometheus.NewGauge(prometheus.GaugeOpts{
+			Name:      fmt.Sprintf("%s_%s_%s", m.Name, m.Tag, key),
+			Namespace: "convey",
+			Subsystem: "convey",
+			Help:      fmt.Sprintf("Convey Metrics %s for %s and value %s", m.Name, m.Tag, key),
+		})
+		gague = m.metrics[key]
+	}
+
+	gague.Inc()
+	return func() { gague.Dec() }, nil
+}
+
+func (m *cMetric) GetMetrics() []prometheus.Collector {
+	metrics := make([]prometheus.Collector, len(m.metrics))
+
+	index := 0
+	for _, v := range m.metrics {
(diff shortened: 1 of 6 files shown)


convey/conveymetric/metric.go
```diff
@@ -0,0 +1,76 @@
...
+
+type cMetric struct {
+	provider provider.Provider
+	tag      string
+	name     string
+	metrics  map[string]metrics.Gauge
+}
+
+func (m *cMetric) Update(data convey.C) (MetricClosure, error) {
+	var gague metrics.Gauge
```
johnabass: it's spelled `gauge` :)

convey/conveymetric/metric.go
```diff
@@ -0,0 +1,76 @@
...
+		name:     name,
+		metrics:  make(map[string]metrics.Gauge),
+	}
+}
+
+type cMetric struct {
+	provider provider.Provider
+	tag      string
+	name     string
+	metrics  map[string]metrics.Gauge
```
johnabass: Don't use a map here.  Use the label API.  Have a single `metric metrics.Gauge` struct member.  Then, use a label for each model:

```
cMetric.metric.With("model", value).Add(1.0)
```

The gauge will then have to be configured with the "model" label.

convey/conveymetric/metric.go
```diff
@@ -0,0 +1,76 @@
...
+	Update(data convey.C) (MetricClosure, error)
+}
+
+var (
+	validKeyRegex   = regexp.MustCompile(`^[a-zA-Z_:]+$`)
+	replaceRegex    = regexp.MustCompile(`[\W]+`)
+	underscoreRegex = regexp.MustCompile(`_+`)
+)
+
+func NewConveyMetric(provider provider.Provider, tag string, name string) CMetric {
```
johnabass: It's best not to depend on the `Provider` interface.  Instead, simply pass a `metrics.Gauge`.  Rely on enclosing application code to configure the provider and create the gauge.  This makes this much more testable, and isolates this code from creational semantics.  Creating a gauge is a nontrivial thing that involves labels, naming conventions, etc.  That's best left outside of library code.

convey/conveymetric/metric.go
```diff
@@ -0,0 +1,50 @@
...
+		tag:   tag,
+		metricName:  metricName,
+		gauge: gauge,
+	}
+}
+
+// cMetric is the internal CMetric implementation
+type cMetric struct {
+	tag   string
+	metricName  string
```
johnabass: This is called a "label" in the metrics APIs, both go-kit and Prometheus.  The metric name is actually the name of the gauge itself (unknown to this code).

convey/conveymetric/metric.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// UnknownLabel is a constant for when key/tag can not be found in the C JSON
+const UnknownLabel = "unknown"
+
+// MetricClosure will be returned after update of struct, this should be used to update the struct, aka decrement the count
+type MetricClosure func()
+
+// CMetric provides
+type CMetric interface {
```
johnabass: This "stutters".  Client code refers to this type as `conveymetric.CMetric`.  Unlike other languages, you can't omit the package name when referring to the type.

It's customary in golang for a central interface in a package that would otherwise share the name of the package to be called `Interface`, e.g. `type Interface interface { ... }`.  Alternatively, you can use a single appropriate letter, e.g. `type C interface {...}`.  I personally prefer `Interface` in this particular case, but that's just a preference.

This concept is discussed at some length here:
https://golang.org/doc/effective_go.html#package-names

convey/conveymetric/metric.go
```diff
@@ -0,0 +1,50 @@
...
+func NewConveyMetric(gauge metrics.Gauge, tag string, label string) CMetric {
+	return &cMetric{
+		tag:   tag,
+		label: label,
+		gauge: gauge,
+	}
+}
+
+// cMetric is the internal CMetric implementation
+type cMetric struct {
```
johnabass: The unexported concrete type that implements the central package interface can be called whatever you like.  This is a fine name :)

Issue #362, opened 2018-11-15
SNS Unsubscribe doesn't check Subscription ID (ARN)


ajgale:
When unsubscribing from old SNS subscriptions, the SubscriptionID (topicARN) is not being checked for invalid values, such as "Deleted" or "PendingConfirmation"

joe94:
Minor but still valid defect. It'll serve to avoid unsubscribing from subscriptions in the "pending confirmation" state. SNS doesn't allow unsubscribing/deleting these as they are pruned after 3 days of age.

joe94:
We're abandoning SNS in favor or Argus (https://github.com/xmidt-org/argus) for storing webhook subscriptions.

PR #363, opened 2018-11-15
Fix SNS unsubscribe error

ajgale:
Associated Issue #362 

When unsubscribing from old SNS subscriptions, the
SubscriptionID (topicARN) was not being checked for
invalid values, such as "Deleted" or "PendingConfirmation"

Also fixes incorrect debug output when there is
an error returned from SNS internal "Unsubscribe"

### DIFF
--- a/webhook/aws/sns_handler.go
+++ b/webhook/aws/sns_handler.go
@@ -348,9 +348,9 @@ func (ss *SNSServer) Unsubscribe(subArn string) {
 
 	if err != nil {
 		ss.errorLog.Log(logging.MessageKey(), "SNS Unsubscribe error", logging.ErrorKey(), err)
+	} else {
+		ss.debugLog.Log(logging.MessageKey(), "Successfully unsubscribed from the SNS topic", "response", resp)
 	}
-
-	ss.debugLog.Log(logging.MessageKey(), "Successfully unsubscribed from the SNS topic", "response", resp)
 }
 
 func (ss *SNSServer) UnsubscribeOldSubscriptions() {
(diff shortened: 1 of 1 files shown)


webhook/aws/sns_handler.go
```diff
@@ -402,7 +402,8 @@ func (ss *SNSServer) ListSubscriptionsByMatchingEndpoint() (*list.List, error) {
 		for _, sub := range resp.Subscriptions {
 			timestamp = 0
 			if !strings.EqualFold(*sub.Endpoint, ss.SelfUrl.String()) &&
-				strings.Contains(*sub.Endpoint, endpoint) {
+				strings.Contains(*sub.Endpoint, endpoint) &&
+				strings.Contains(*sub.SubscriptionArn, ss.Config.Sns.TopicArn) {
```
shilpa24balaji: Any sample example data to prove this condition?

ajgale: Unfortunately I don't have sample data and don't have the time to recreate the issue on my end. 

I remember we were having an issue with Caduceus confirming its subscriptions, which caused us to have multiple "Pending confirmation" subs on the AWS SNS console for that particular topic. These would be included when requesting current subscriptions. To recreate you should first send some subscription events without confirming them, then restarting the server.

joe94:
We're dropping dependency on SNS for webhooks. We plan to use our own registry datastore. Progress for the work can be followed here: https://github.com/orgs/xmidt-org/projects/2?card_filter_query=label%3Aargus-webhooks

PR #366, opened 2018-11-19
add bookkeeping library

kcajmagic:


bookkeeping/bookkeeper.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// RequestFunc takes the request and returns key value pairs from the request
+type RequestFunc func(request *http.Request) []interface{}
+
+// ResponseFunc takes the ResponseWriter and returns key value pairs from the request
+type ResponseFunc func(response CapturedResponse) []interface{}
+
+// Handler is for logging of the request and response
+type Handler struct {
```
johnabass: This shouldn't be exported from the package, given that you have a `New` constructor.  This prevents a lot of confusion over how the package's code should be used.

bookkeeping/bookkeeper.go
```diff
@@ -0,0 +1,123 @@
...
+// WithResponses take a one or many ResponseFunc to build an Option for logging key value pairs from the
+// ResponseFunc
+func WithResponses(responseFuncs ...ResponseFunc) Option {
+	return func(h *Handler) {
+		h.after = append(h.after, responseFuncs...)
+	}
+}
+
+// WithLogger will set a custom logger for the Bookkeeping Handler
+func WithLogger(logger log.Logger) Option {
```
johnabass: Don't supply a logger via configuration.  All HTTP code should obtain the logger from the request's context, like so:

```
logging.GetLogger(request.Context()).Log(...)
```

There is an implicit default supplied by our `logging` package, so `GetLogger` never returns `nil`.

bookkeeping/bookkeeper.go
```diff
@@ -0,0 +1,123 @@
...
+func (h *Handler) ServeHTTP(responseWriter http.ResponseWriter, request *http.Request) {
+	kv := []interface{}{logging.MessageKey(), "Bookkeeping transactor"}
+	for _, before := range h.before {
+		kv = append(kv, before(request)...)
+	}
+
+	w := &writerInterceptor{ResponseWriter: responseWriter}
+
+	start := time.Now()
+	h.next.ServeHTTP(w, request)
```
johnabass: The bookkeeping cleanup should generally be inside a `defer` call.  This allows bookkeeping to be done on code that panics.

bookkeeping/bookkeeper.go
```diff
@@ -0,0 +1,123 @@
...
+		Payload: w.buffer.Bytes(),
+		Header:  w.Header(),
+	}
+
+	for _, after := range h.after {
+		kv = append(kv, after(response)...)
+	}
+
+	var logger log.Logger
+	if h.logger != nil {
```
johnabass: See my comment above about the contextual logger.  We don't want to have HTTP code explicitly configured with a logger.

bookkeeping/bookkeeper.go
```diff
@@ -0,0 +1,123 @@
...
+		logger = h.logger
+	} else {
+		logger = logging.GetLogger(request.Context())
+	}
+
+	logging.Info(logger).Log(kv...)
```
johnabass: Don't forget that the `logginghttp` package can insert key/value pairs for common request properties, such as the request URI.  In general, common request properties shouldn't be handled by this code.  Allow a client to set those bookkeeping items up outside this package.  This allows all logging on a particular HTTP codepath to see these key/value pairs.

PR #369, opened 2018-11-28
Wire money webpa transactors

githubsands:
Wires money into transactors necessary for servicehttp and fanout


githubsands:
#355  closes this

PR #372, opened 2018-12-05
wrp: include money spans in new wrp messages

githubsands:
1. as intended by the rdk user story - "Add timing collection and reporting to parados"
2. include new middleware for talaria's device handlers that decode and encode


githubsands:
#355 closes this

PR #400, opened 2019-04-25
server: add configuration for tlsconfig min and max values

githubsands:


githubsands:
#401  solves this - branch error due to new machine.

Issue #403, opened 2019-04-30
xmetrics unit test is fails sometimes


schmidtw:
It looks like the test code doesn't take into consideration that the JSON encoding of the order can change.

```
--- FAIL: TestReturnHeadersWithPrefix (0.06s)
    --- FAIL: TestReturnHeadersWithPrefix/10 (0.04s)
        requestResponse_test.go:170: struct { request *http.Request; prefixs []string; expectedKV []interface {} }{request:(*http.Request)(0xc000189d00), prefixs:[]string{"X-TEST"}, expectedKV:[]interface {}{"X-Test-1", []string{"foo"}, "X-Test-2", []string{"foo", "bar"}}}
	Error Trace:	requestResponse_test.go:102
			requestResponse_test.go:171
	Error:      	Not equal: 
	            	expected: []interface {}{"X-Test-1", []string{"foo"}, "X-Test-2", []string{"foo", "bar"}}
	            	actual  : []interface {}{"X-Test-2", []string{"foo", "bar"}, "X-Test-1", []string{"foo"}}
	            	
	            	Diff:
	            	--- Expected
	            	+++ Actual
	            	@@ -1,6 +1,2 @@
	            	 ([]interface {}) (len=4) {
	            	- (string) (len=8) "X-Test-1",
	            	- ([]string) (len=1) {
	            	-  (string) (len=3) "foo"
	            	- },
	            	  (string) (len=8) "X-Test-2",
	            	@@ -9,2 +5,6 @@
	            	   (string) (len=3) "bar"
	            	+ },
	            	+ (string) (len=8) "X-Test-1",
	            	+ ([]string) (len=1) {
	            	+  (string) (len=3) "foo"
	            	  }
	Test:       	TestReturnHeadersWithPrefix/10
```

githubsands:
#406 solves this.

PR #404, opened 2019-05-01
bookkeeping: fix requestReponse_test.go

githubsands:
Test `requestResponse_test.go` does not take into account how gos map's hash function distributes keys suedo randomly within the hash table.


githubsands:
solves #403

PR #408, opened 2019-05-09
device: migrate to comcast/wrp-go

githubsands:
Moves device package from webpa-common to wrp-go.

### DIFF
--- a/device/device_test.go
+++ b/device/device_test.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/Comcast/webpa-common/logging"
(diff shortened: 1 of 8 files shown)


kristinaspring:
Hi Ryan, I don't think we should move forward with this until we get rid of webpa-common dependencies in wrp-go:
https://github.com/Comcast/wrp-go/search?q=webpa-common&unscoped_q=webpa-common

The plan is, once these dependencies are taken care of, to release a new major version of wrp-go that we can use in webpa-common.

Issue #62, opened 2017-03-03
Claims Validation Error


schmidtw:
If you pass in a nasty claim you can bypass the protection.

`/api/v2/device/mac:112233445566/stat?ignore-me=/api/v2/device/mac:112233445566/hook`
or
`/api/v2/device/mac:112233445566/stat/../../../../../api/v2/device/mac:112233445566/hook`

ilawjr:
This issue is still valid.

kristinaspring:
Closing this issue, but will investigate and possibly open a new one in the bascule repo.

kristinaspring:
This is still an issue in the `basculechecks` package:
https://github.com/xmidt-org/webpa-common/blob/master/basculechecks/basculechecks.go#L59

instead of matching anywhere in the string, we need to be more judicious about how we are matching, to ensure the requester is authorized to access the endpoint.

kristinaspring:
The first case is dealt with from using the URL EscapedPath() function, which doesn't include parameters:
https://golang.org/pkg/net/url/#URL.EscapedPath

Still working on a fix for the second case and making our regex more picky.

kristinaspring:
Going to see what more needs to be done for this issue, if anything.

kristinaspring:
This issue has been fixed.

PR #440, opened 2019-11-04
Modified capabilities check to be more restrictive

kristinaspring:
fixes part of https://github.com/xmidt-org/webpa-common/issues/62

### DIFF
--- a/basculechecks/basculechecks.go
+++ b/basculechecks/basculechecks.go
@@ -13,7 +13,6 @@ var (
 	ErrNoVals                 = errors.New("expected at least one value")
 	ErrNoAuth                 = errors.New("couldn't get request info: authorization not found")
 	ErrNonstringVal           = errors.New("expected value to be a string")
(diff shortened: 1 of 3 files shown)


JC000:
Looks OK to me other than the unit test failure.

basculechecks/basculechecks.go
```diff
@@ -14,18 +13,11 @@ var (
...
-type CapabilityConfig struct {
-	FirstPiece      string
-	SecondPiece     string
-	ThirdPiece      string
-	AcceptAllMethod string
-}
-
-func CreateValidCapabilityCheck(config CapabilityConfig) func(context.Context, []interface{}) error {
+func CreateValidCapabilityCheck(prefix string, acceptAllMethod string) func(context.Context, []interface{}) error {
+	matchPrefix := regexp.MustCompile("^" + prefix + "(.+):(.+?)$")
```
kcajmagic: this should be error checked,  or some modification/validation should happen to `prefix`

PR #447, opened 2020-01-17
Device Metadata

joe94:
### Changes introduced:
- Replace direct getter/setters with a single metadata map object which defines the interface of a device metadata.
- Expand device/context.go for a device metadata
- Have context objects as first arguments in device/context.go functions
- No longer supporting SATClientID on a device which included the SAT JWT token passed when the device connected. If we really need it, we'd need to either have a custom bascule constructor or enrich a bascule token to have the full source http authorization header


device/device_metadata.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (m Metadata) SetJWTClaims(jwtClaims JWTClaims) {
+	m[JWTClaimsKey] = jwtClaims
+}
+
+func (m Metadata) SessionID() string {
+	return m[SessionIDKey].(string)
+}
+
+// Load allows retrieving values from a device's metadata
```
kristinaspring, line 49:
```suggestion
// Load allows retrieving values from a device's metadata.
```

device/device_metadata.go
```diff
@@ -0,0 +1,117 @@
...
+// Note: operations will fail for reserved keys.
+func (m Metadata) Store(key string, value interface{}) bool {
+	if reservedMetadataKeys[key] {
+		return false
+	}
+	m[key] = value
+	return true
+}
+
+// NewDeviceMetadata returns a metadata object ready for use
```
kristinaspring, line 65:
```suggestion
// NewDeviceMetadata returns a metadata object ready for use.
```

device/device_metadata.go
```diff
@@ -0,0 +1,117 @@
...
+func NewDeviceMetadata() Metadata {
+	m := make(Metadata)
+	m.SetJWTClaims(NewJWTClaims(make(map[string]interface{})))
+	m[SessionIDKey] = ksuid.New().String()
+	return m
+}
+
+// NewJWTClaims is a convenience constructor useful for setting
+// claims on an existing device metadata object
```
kristinaspring, line 74:
```suggestion
// claims on an existing device metadata object.
```

device/device_metadata.go
```diff
@@ -0,0 +1,117 @@
...
+func NewJWTClaims(claims map[string]interface{}) JWTClaims {
+	return JWTClaims{
+		data: claims,
+	}
+}
+
+// JWTClaims defines the allowed interactions with the claims
+// in a device's metadata
```
kristinaspring, line 82:
```suggestion
// in a device's metadata.
```

device/device_metadata.go
```diff
@@ -0,0 +1,117 @@
...
+type JWTClaims struct {
+	data map[string]interface{}
+}
+
+// Data returns the internal claims map
```
kristinaspring, line 87:
```suggestion
// Data returns the internal claims map.
```

device/device_metadata.go
```diff
@@ -0,0 +1,117 @@
...
+func (c *JWTClaims) Data() map[string]interface{} {
+	return c.data // TODO: return a deep copy if security/mutability is a concern?
+}
+
+// Trust returns the device's trust level claim
```
kristinaspring, line 92:
```suggestion
// Trust returns the device's trust level claim.
```

device/device_metadata.go
```diff
@@ -0,0 +1,117 @@
...
+// By Default, a device is untrusted (trust = 0).
+func (c *JWTClaims) Trust() int {
+	if trust, ok := c.data[TrustClaimKey].(int); ok {
+		return trust
+	}
+	return 0
+}
+
+// PartnerIDs returns a singleton list with the partnerID
+// the device presented during registration
```
kristinaspring, line 102:
```suggestion
// the device presented during registration.
```

device/manager.go
```diff
@@ -149,41 +148,32 @@ func (m *manager) Connect(response http.ResponseWriter, request *http.Request, r
...
+		ID:         id,
+		C:          cvy,
+		Compliance: convey.GetCompliance(cvyErr),
+		QueueSize:  m.deviceMessageQueueSize,
+		Metadata:   metadata,
+		Logger:     m.logger,
 	})
 
+	if !providedDeviceMetadata {
+		d.errorLog.Log(logging.MessageKey(), "missing security information")
```
kristinaspring, line 168: Can't this go in the if statement above? They're both checking the same thing.

device/metadata.go
```diff
@@ -0,0 +1,138 @@
...
+
+// JWTClaims defines the intended interactions with the claims
+// in a device's metadata. The main current use case includes
+// having this object represent the claims presented by devices
+// during registration to the XMiDT cluster.
+type JWTClaims map[string]interface{}
+
+// ToMap returns the casted map representation of the claims.
+// Note: Return value is not a copy so caution is advised.
+func (c JWTClaims) ToMap() map[string]interface{} {
```
joe94, line 107: We don't actually need this anymore. Just use direct casting.

device/metadata.go
```diff
@@ -0,0 +1,138 @@
...
+// given claims.
+func NewDeviceMetadataWithClaims(claims map[string]interface{}) Metadata {
+	m := make(Metadata)
+	m.SetJWTClaims(JWTClaims(claims))
+	m.initSessionID()
+	return m
+}
+
+// JWTClaims defines the intended interactions with the claims
+// in a device's metadata. The main current use case includes
```
joe94, line 100: Fix circular description

device/manager.go
joe94: I think this is done so we can use the device's logger

kristinaspring: I'm talking about the if statement, not the log message.  We check the same parameter (`!providedDeviceMetadata`) in two separate if statements that each have one line in them:
https://github.com/xmidt-org/webpa-common/pull/447/files#diff-540e973a8df442549b241f323503fb38R167
and
https://github.com/xmidt-org/webpa-common/pull/447/files#diff-540e973a8df442549b241f323503fb38R153

joe94: The second if condition was actually not sufficient for its purspose. I changed it https://github.com/xmidt-org/webpa-common/pull/447/commits/42263b806f0c2c8e6c75741921ff21b8fa418a1b

joe94: But I'm going through to ensure it does what I think it should

PR #451, opened 2020-01-31
add session-id to event metadata

kcajmagic:


device/device.go
```diff
@@ -180,6 +185,7 @@ func newDevice(o deviceOptions) *device {
 		transactions: NewTransactions(),
 		partnerIDs:   partnerIDs,
 		satClientID:  o.SatClientID,
+		sessionID:    ksuid.New().String(),
```
johnabass: This is the only place `ksuid` is used, and generating a UUID is very simple:

```
var id [16]byte
rand.Read(id[:])
sessionID := base64.RawURLEncoding.EncodeToString(id[:]) // or however you want to encode it
// URL-safe encoding is usually best, plus base64 is smaller than simply hexadecimal
```

That avoids another dependency (and any of its transitive dependencies).

PR #454, opened 2020-02-06
add session id to request response

kcajmagic:


device/manager.go
```diff
@@ -321,6 +321,22 @@ func (m *manager) readPump(d *device, r ReadCloser, closeOnce *sync.Once) {
 			m.measures.RequestResponse.Add(1.0)
 		}
 
+		// update message metadata fields
+		if message.Metadata == nil {
+			message.Metadata = map[string]string{}
+		}
+		message.Metadata["partner-ids"] = strings.Join(event.Device.PartnerIDs(), ",")
```
johnabass: `Device.PartnerIDs()` came from the device itself, at connection time.  If the device didn't send a partner id then, it wouldn't be at this point.  This doesn't help the missing partner id problem.

If we want to ensure the partner ids match, which would only be due to a device bug, then we should set `message.PartnerIDs` to the device's partner ids.  Again, I'm not sure what the value of that is.

device/manager.go
```diff
@@ -321,6 +321,22 @@ func (m *manager) readPump(d *device, r ReadCloser, closeOnce *sync.Once) {
...
+		message.Metadata["session-id"] = event.Device.SessionID()
```
johnabass: Do we really need this?  Remember that metadata is supposed to be for a device's RDK stack to populate.  We use this for connect/disconnect events in talaria as hack, since those events don't come from devices.

PR #462, opened 2020-04-02
Abandon racy testLogger using testing.T or testing.B

joe94:
Using the `Log(...)` function of testing.T or testing.B  as a delegate for our testLogger has introduced intermittent races across our test suite. Since there doesn't seem to be an obvious fix to working with it, we're opting on printing directly to stdout when verbose testing is on.

Additionally, added some minor improvements to our travis ci config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.13.x
+  - 1.14.x
   - tip
 
 os:
@@ -17,13 +17,16 @@ before_install:
   - chmod 755 ./codecov.sh
 
 install:
-  - GO111MODULE=on go mod vendor
+  - go mod vendor
 
 script:
-  - go test ./... -race -coverprofile=coverage.txt 
+  - go test ./... -v -race -coverprofile=coverage.txt 
   - ./codecov.sh
 
 jobs:
+  fast_finish: true
+  allow_failures:
+    - go: tip
   include:
     - stage: tag
       name: "Tag For Release"
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,11 @@ install:
   - go mod vendor
 
 script:
-  - go test -v -race -coverprofile=coverage.txt ./...
+  # Read https://github.com/xmidt-org/webpa-common/pull/462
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.13.x
+  - 1.14.x
```
johnabass, line 4: Are we officially on `go1.14` yet?

logging/testLogger.go
```diff
@@ -12,23 +16,29 @@ type testSink interface {
 	Log(...interface{})
 }
 
-// testWriter implements io.Writer and delegates to a testSink
+// testWriter implements io.Writer
```
johnabass, line 19: go-kit already has a type for exactly this purpose:
https://pkg.go.dev/github.com/go-kit/kit/log?tab=doc#NewSyncWriter

.travis.yml
joe94: All of our other services have moved CI go version to `go1.14`. copr is holding us back though for building RPMs with this version  

Unless it lets us install go from another registry, https://dl.fedoraproject.org/pub/epel/7/x86_64/Packages/g/ only has `go1.13.6` as latest.

logging/testLogger.go
johnabass: You actually don't need to synchronize the `NopLogger`.  Just returning `NewNopLogger` is sufficient.

logging/testLogger.go
```diff
@@ -12,35 +12,24 @@ type testSink interface {
...
-	return &testWriter{testSink: t}
-}
+// NewTestLogger produces a go-kit Logger which logs to stdout only if
+// the verbose testing mode.
+// Note: Although originally intended to delegate data to testSink,
+// intermittent data races have forced us to stick to writing directly
+// to stdout and do the verbose check outselves.
+func NewTestLogger(o *Options, _ testSink) log.Logger {
+	if !testing.Verbose() {
+		return log.NewSyncLogger(log.NewNopLogger())
```
johnabass, line 22: Just `log.NewNopLogger` is sufficient.  It doesn't need synchronization.

PR #480, opened 2020-05-04
Add filtering capability for rehasher and fix sd metric labels

joe94:


device/rehasher/rehasher.go
```diff
@@ -116,6 +123,7 @@ func New(connector device.Connector, options ...Option) monitor.Listener {
 // and (2) rehashes devices in response to updated instances.
 type rehasher struct {
 	logger          log.Logger
+	services        []string
```
johnabass, line 126: Even though the number of services is small (or even just one), it's still clearer to make this a `map[string]bool`.  That's the golang idiom for a "set" of things.  You can then use zero values for set inclusion with `services["foo"]`.

PR #483, opened 2020-05-16
Bugfix/jwt claims

joe94:


device/metadata.go
```diff
@@ -34,9 +35,9 @@ type Metadata map[string]interface{}
 // they are initialized appropiately.
 func (m Metadata) JWTClaims() JWTClaims { // returns the type and such type has getter/setter
```
johnabass, line 36: Using copy-on-read is fine here.  However, another possible approach that avoids copying at all is to use a predicate to access the claims.  I'm not sure if this will play well with the existing codebase, though.

Here's an example:

```
// assume Metadata has a field like "lock sync.Mutex"
func (m Metadata) UseJWtClaims(f func(JWTClaims) error) error {
    m.lock.Lock()
    err := f(m[JWTClaimsKey].(JWTClaims))
    m.lock.Unlock()
    return err
}
```

joe94: Unfortunately, this won't play nice with the viper integration which requires a copy. Moving the claims getters/setters up to the metadata + synchronization could help avoid unnecessary copies though.

Issue #422, opened 2019-09-17
Add function to create bascule middleware


kristinapathak:
These services all have a similar function that takes some input and makes an Alice.Chain from it.  It would be nice to have a generic helper functions that all services can use:
https://github.com/xmidt-org/gungnir/blob/57b487331cde31610068bf52634e948700560161/primaryHandler.go#L302
https://github.com/xmidt-org/scytale/blob/b6600e012b6910f5861bcecc0dd84eb5adc014fa/primaryHandler.go#L76
https://github.com/xmidt-org/tr1d1um/blob/b117f90bda5d29cea5283da3abee1165cfbf4921/main.go#L394

PR #487, opened 2020-05-26
Basculecheck newfunc

kristinaspring:
part of https://github.com/xmidt-org/webpa-common/issues/422

also added some comments.

### DIFF
--- a/basculechecks/basculechecks.go
+++ b/basculechecks/basculechecks.go
@@ -37,6 +37,11 @@ type capabilityCheck struct {
 
 var defaultLogger = log.NewNopLogger()
 
+// CreateBasculeCheck creates a function that determines whether or not a
+// client is authorized to make a request to an endpoint by comparing the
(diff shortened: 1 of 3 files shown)


basculechecks/basculechecks.go
```diff
@@ -107,6 +117,30 @@ func NewCapabilityChecker(m *AuthCapabilityCheckMeasures, prefix string, acceptA
 	return &c, nil
 }
 
+// NewCapabilityCheckerFromStrings creates the capability checker, and allows
+// consumers to provide a list of string endpoints, rather than regular
```
joe94, line 121: Maybe rephrase to something like: 
```
// NewCapabilityCheckerFromStrings creates the capability checker from a list of endpoint regular expressions as 
// strings to be compiled. Only expressions that compile successfully are included in the checker.
```
To be explicit about how the strings will be compiled to a re and that only the successfully compiled ones will make the final check list.

PR #502, opened 2020-07-21
renamed AllDataCenters to WatchAllByDatacenter

kcajmagic:
This separates out any confusion on how the list of instances appear in the MonitorEvent function. We may want to add AllDatacenters back for when we want the list of all nodes across all datacenters.

### DIFF
--- a/service/consul/environment.go
+++ b/service/consul/environment.go
@@ -130,7 +130,7 @@ func newInstancers(l log.Logger, c Client, co Options) (i service.Instancers, er
 	var datacenters []string
 
 	for _, w := range co.watches() {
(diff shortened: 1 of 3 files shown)


service/consul/environment.go
```diff
@@ -69,9 +69,6 @@ func ensureIDs(r *api.AgentServiceRegistration) {
 
 func newInstancerKey(w Watch) string {
 	datacenter := w.QueryOptions.Datacenter
```
johnabass, line 71: we should remove this and put `w.QueryOptions.Datacenter` directly into the `fmt.Sprintf` call below

PR #503, opened 2020-07-23
copy scheme for fanout

kcajmagic:


joe94:
Discussed with jack and this didn't solve what was hoped for so it's NLR.

PR #507, opened 2020-07-31
Validate WRP.Source for inbound messages

joe94:
The main logic introduced is summarized here: https://github.com/xmidt-org/webpa-common/pull/507/files#diff-4908257ccaf67982253a9806f535dd1dR75

### DIFF
--- a/device/options.go
+++ b/device/options.go
@@ -27,6 +27,10 @@ const (
 	DefaultDeviceMessageQueueSize = 100
 )
 
+type wrpSourceCheckConfig struct {
+	Type string
+}
+
 // Options represent the available configuration options for components
 // within this package
 type Options struct {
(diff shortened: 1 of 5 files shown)


JC000:
As long as it isn't some sort of coding error on our part, there shouldn't be a device identifier that is *different* between a message (simple event message or otherwise) source and the device identifier of the "connection" (connection context).  The only possible exception is an empty source.  I think it's OK to populate an empty source field, but a different source is an indication of something malicious.  Log a security exception and drop the message on the floor.

device/options.go
```diff
@@ -27,6 +27,10 @@ const (
 	DefaultDeviceMessageQueueSize = 100
 )
 
+type wrpSourceCheckConfig struct {
```
kcajmagic, line 30: why is this private?

joe94: We don't need it elsewhere outside the package. We do need the "Type" to be exported for viper to work properly.

PR #509, opened 2020-08-04
Add eventKey to service discovery metric 

kcajmagic:
The metric would be wrong if the flag `crossDatacenter` was enabled.


kcajmagic:
required for xmidt-org/scytale#109

PR #512, opened 2020-08-05
Add trust label to hardware_model gauge

joe94:


device/metrics.go
```diff
@@ -57,7 +57,7 @@ func Metrics() []xmetrics.Metric {
 		{
 			Name:       ModelGauge,
 			Type:       "gauge",
-			LabelNames: []string{"model", "partnerid", "firmware"},
+			LabelNames: []string{"model", "partnerid", "firmware", "trust"},
```
kristinaspring, line 60: I think we should change these label name strings to `const`s so wherever they are referenced with metrics uses the constant instead of having the same string written out in multiple places.  This would make it much harder for someone to add to a metric somewhere and typo a label name, and it would make it easier to change a label name in the future.

But this is the standard in talaria, so I don't think that kind of change belongs in this pr.

joe94: ➕  I'd be on board with that approach

PR #518, opened 2020-08-25
Metrics and Typo fixes

joe94:


device/options.go
```diff
@@ -9,6 +9,12 @@ import (
 	"github.com/xmidt-org/webpa-common/logging"
 )
 
+// Check types for the WRP Source check
+const (
```
johnabass, line 13: As it is now, `CheckTypeMonitor` and `CheckTypeEnforce` are just constants that can be used anywhere a `string` is used.  Most of the time, that's not what you want.  The compiler can't complain if someone passes those constants where they shouldn't.

```
func NothingToDoWithMetrics(v string) { /* stuff */ }

// this works, but should it?
// is it what you intended, and/or would it surprise the consumer of your API?
NothingToDoWithMetrics(CheckTypeMonitor)
```

Also, someone can pass any old string like "bob's your uncle" to your own code that is expecting one of the CheckType values.

A better approach is a custom type, possibly enumerated:

```
// Variant 1: just a custom string type
type CheckType string
const (
    CheckTypeMonitor CheckType = "monitor"
    CheckTypeEnforce CheckType = "enforce"
)

// Variant 2: enumerated type that implements fmt.Stringer.
type CheckType int
const (
    CheckTypeMonitor CheckType = iota // zero value, great for defaults
    CheckTypeEnforce
)

func (ct CheckType) String() { /* return the appropriate string */ }
```

Both variants help constrain things so that your code clearly specifies what it requires.  It also warns a programmer if they use a constant where it doesn't make sense.  You have to do an explicit cast to get around this, making the intent very clear.

```
var ct CheckType
NothingToDoWithMetrics(ct) // this will now break, requiring the programmer to explicitly cast

// You can still pass a CheckType, but you have to be explicit that you want that
NothingToDoWithMetrics(string(ct)) // variant 1
NothingToDoWithMetrics(ct.String()) // variant 2
```

joe94: Good idea! I like variant 1 as it's a bit more concise 👍

PR #522, opened 2020-09-17
Basculechecks

kristinaspring:
Splitting up the capability check into multiple structs wrapping each other, for separating metrics from the checking logic and allowing for different checks against capabilities.


basculechecks/capabilitycheck.go
```diff
@@ -0,0 +1,99 @@
...
+	"strings"
+
+	"github.com/goph/emperror"
+)
+
+var (
+	ErrNilPrefix = errors.New("prefix regular expression cannot be nil")
+)
+
+type constCheck struct {
```
johnabass, line 32: This is a good approach, with a single interface being implemented by several concrete types.

However, since you're already encapsulating things with an interface, you could just have a plain, exported types and forgo any `New...` function:

```
type ConstCheck string

func (c ConstCheck) Authorized(....) { /* stuff */ }

type EndpointRegexCheck struct {
    PrefixToMatch *regexp.Regexp
    AcceptAllMethod string
}

func (c EndpointRegexCheck) Authorized(...) { /* stuff */ }

// etc etc for the other types
```

That all takes up less code without really giving up any information hiding.  It also gives you one less thing per type to maintain, as if you add fields to something you don't need to update a `New...` function.  You just add the field to the struct and modify `Authorized`.

basculechecks/metricvalidator.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	mc := metricValidator{
+		measures:  m,
+		c:         c,
+		endpoints: endpoints,
+	}
+	return &mc, nil
+}
+
+func NewMetricCapabilityCheckerFromStrings(c CapabilitiesChecker, m *AuthCapabilityCheckMeasures, endpoints []string, logger log.Logger) (*metricValidator, error) {
```
johnabass, line 109: This function is an example of what I was getting at earlier.  It's signature is kinda lengthy, and over time it may grow with even more parameters.  If you just export `MetricValidator` and export the relevant fields, it'll save you having to maintain this function.

As long as you write all your code internally to the `CapabilitiesChecker` interface, you've accomplished good encapsulation.  Additionally, code outside `bascule` can be implemented in terms of `CapabilitiesChecker`.  So, the only code that needs to be aware of the specific struct and its fields is the point where its created.

basculechecks/capabilitiesvalidator.go
```diff
@@ -0,0 +1,115 @@
...
+type capabilitiesValidator struct {
+	checker CapabilityChecker
+}
+
+// CreateBasculeCheck creates a function that determines whether or not a
+// client is authorized to make a request to an endpoint by comparing the
+// method and url to the values at the CapabilityKey in the Attributes of a
+// token.  The function created can error out or not based on the parameter
+// passed, and the outcome of the check will be updated in a metric.
+func (c *capabilitiesValidator) CreateBasculeCheck(errorOut bool) bascule.ValidatorFunc {
```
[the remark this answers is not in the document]
kristinaspring: Oops, meant to rename this to `CreateValidator()`.  This isn't a CapabilityChecker (the interface) or what makes the validator a CapabilitiesChecker (a different interface), so I don't want to call the function it's creating a capability check.

basculechecks/capabilitiesvalidator.go
```diff
@@ -0,0 +1,115 @@
...
+		return NoCapabilitiesMatch, err
+	}
+	return "", nil
+}
+
+// NewCapabilitiesParser creates an object that produces a check on capabilities
+// in bascule tokens, to be run by the bascule enforcer middleware.
+func NewCapabilitiesParser(c CapabilityChecker) (*capabilitiesValidator, error) {
+	if c == nil {
+		return nil, ErrNilPrefix
```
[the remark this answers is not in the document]
kristinaspring: Whoops, good catch. :smile:

basculechecks/capabilitiesvalidator.go
```diff
@@ -85,15 +95,13 @@ func (c *CapabilitiesValidator) Check(auth bascule.Authentication) (string, erro
 
 // checkCapabilities parses each capability to check it against the prefix
 // expected, the url hit, and the method used.  If a match is found, no error is returned.
-func (c *CapabilitiesValidator) checkCapabilities(capabilities []string, requestInfo bascule.Request) error {
-	urlToMatch := requestInfo.URL.EscapedPath()
-	methodToMatch := requestInfo.Method
+func (c CapabilitiesValidator) checkCapabilities(capabilities []string, reqURL string, method string) error {
```
joe94, line 98: Good call on simplifying the args for this func 👍

basculechecks/capabilitycheck.go
```diff
@@ -29,23 +30,42 @@ func (c ConstCheck) Authorized(capability, _, _ string) bool {
 }
 
 type EndpointRegexCheck struct {
-	PrefixToMatch   *regexp.Regexp
-	AcceptAllMethod string
+	prefixToMatch   *regexp.Regexp
+	acceptAllMethod string
+}
+
+// NewCapabilityChecker creates an object that produces a check on capabilities
```
joe94, line 37: Comment not matching function

joe94 (changes requested):

basculechecks/capabilitycheck.go
```diff
@@ -23,19 +23,29 @@ import (
...
 }
 
-// NewCapabilityChecker creates an object that produces a check on capabilities
-// in bascule tokens, to be run by the bascule enforcer middleware.
+// NewEndpointRegexCheck creates an object that implements the
+// CapabilityChecker interface.  It takes a prefix that is expected at the
+// beinning of a capability and a string that, if provided in the capability,
+// authorizes all methods for that endpoint.  After the prefix, the
+// EndpointRegexCheck expects there to be an endpoint regular expression and an
+//http method - separated by a colon.
```
joe94, line 48:
```suggestion
// NewEndpointRegexCheck creates an object that implements the
// CapabilityChecker interface.  It takes a prefix that is expected at the
// beginning of a capability and a string that, if provided in the capability,
// authorizes all methods for that endpoint.  After the prefix, the
// EndpointRegexCheck expects there to be an endpoint regular expression and an
// http method - separated by a colon (example capability format: `[prefix][regex]:[method]`).
```

PR #523, opened 2020-09-29
Capcheck

kristinaspring:
Made function for determining partner id metric label value to be exported - allowing for other partner ID metrics to use it.  Example of where it can be used is in Svalinn for this new metric: https://github.com/xmidt-org/svalinn/pull/136

Also includes unit tests I somehow failed to commit in my last pr.

Would release the capability check changes.


joe94 (approved):
🥳

kcajmagic (approved):

j-mai (approved):

PR #529, opened 2020-10-10
Do not generate certificate when reading from ClientCACertFile fails

### LINKED ISSUE #467: webpa.go should never generate any certs
This logging error: 
{"caller":"webpa.go:277","error":"open : no such file or directory","level":"error","msg":"Error in reading ClientCACertFile ","ts":"2020-04-03T08:39:53.077214405Z"}

Should be info or be made clearer as if the server can't find ClientCACertFile it will generate its own - which is expected behavior. 

The error message is from here:
https://github.com/xmidt-org/webpa-common/blob/e1656cf32ed996f4f52d24699f75ceb1e885b480/server/webpa.go#L277

j-mai:
Fixes #467


server/webpa_test.go
```diff
@@ -1,9 +1,11 @@
 package server
 
 import (
+	"crypto/tls"
 	"errors"
+
 	//	"github.com/xmidt-org/webpa-common/health"
```
kristinaspring, line 7:
```suggestion
```
While you're here, might as well get rid of this unused import.

kristinaspring (approved):

johnabass (approved):
Looks great!

PR #530, opened 2020-10-15
Periodic watch

### LINKED ISSUE #536: Ambiguous import issue
There seems to be a problem with hashicorp 1.7.0 that results in an ambiguous import error:

```
github.com/xmidt-org/webpa-common/service/consul imports
        github.com/hashicorp/consul/api: ambiguous import: found package github.com/hashicorp/consul/api in multiple modules:
        github.com/hashicorp/consul v1.4.2 
        github.com/hashicorp/consul/api v1.7.0
```

Workaround is to add this line to `go.mod`:
`github.com/hashicorp/consul v1.7.0`

Make sure to not do `go mod tidy` afterwards because the line will go away if you do.

j-mai:
This PR allows for a periodic watch to update the Instancers of a consul environment

Closes xmidt-org/scytale#112 , Fixes #536


service/consul/environment.go
```diff
@@ -209,13 +208,72 @@ func NewEnvironment(l log.Logger, registrationScheme string, co Options, eo ...s
...
+		currentInstancers := e.Instancers()
+		keys := make(map[string]bool)
+		instancersToAdd := make(service.Instancers)
+
+		var datacenters []string
+		var err error
+		for _, w := range co.watches() {
+			if w.CrossDatacenter {
+				if len(datacenters) == 0 {
+					datacenters, err = getDatacenters(l, client, co)
```
joe94, line 247: I think we'd want to get a fresh copy every time in case a new dc was added.

service/environment.go
```diff
@@ -144,6 +147,26 @@ func (e *environment) Instancers() Instancers {
...
+	}
+
+	// remove outdated instancers
+	for key := range e.instancers {
+		if _, ok := currentKeys[key]; !ok {
+			i, found := e.instancers.Get(key)
+
+			if found {
+				i.Stop()
+				delete(e.instancers, key)
```
joe94, line 163: Since the environment returns a copy of the instancers on read https://github.com/xmidt-org/webpa-common/blob/main/service/environment.go#L144 we might run into a data race here. 

We'll have to either synchronize reads/writes or look to see if we could replace the whole instancers map with the new one (in which case we could use https://golang.org/pkg/sync/atomic/#example_Value_config for atomic map swap)

joe94 (commented):
If I understand correctly, these changes to the environment would then be used to update the monitor which Scytale uses to update its internal list of talaria machines for fanout.

So, this approach would allow scytale to update its datacenter list on an interval. Given our context and our experience with gate and drain, I wonder if it'd be better to follow the same standard and have a manual trigger for this?

service/consul/environment.go
j-mai: It was my understanding that `getDatacenters` gets the datacenters that are part of the consul client's catalog, which is updated whenever a datacenter is added or removed right?

kcajmagic: @j-mai  is correct `getDatacenters` returns all the current datacenters.

joe94: So wouldn't we want to make that call at every tick of `createInstancesAgain`? Say updating line 242 to 
```datacenters, err := getDatacenters(l, client, co)```
I just didn't quite understand the logic behind line 246.

Comment (j-mai) [reply service/consul/environment.go:247]: Yeah, that makes sense. I think my reasoning behind adding line 246 is in case `getDatacenters` errors, then maybe we want to try again on the next watch that we are looping through. But I think it does make more sense to just get it at line 242 and if it errors, wait until `createInstancesAgain` ticks again to check for the datacenters again, since it's likely that getting the datacenters on the next watch will just error again.

Comment (joe94) [reply service/consul/environment.go]: Exactly, it'll keep the code simple and readable 😃  And if we see the need to add retries (unless the call already performs retries under the hood) later on, we can definitely do that

**Review by kcajmagic — APPROVED** [review]

File: service/consul/environment.go [hunk 3]
```diff
@@ -209,13 +208,69 @@ func NewEnvironment(l log.Logger, registrationScheme string, co Options, eo ...s
...
+			)...), NewClient(consulClient)}
+
+	if co.DatacenterWatchInterval > 0 {
+		go WatchInstancers(l, co, newServiceEnvironment)
+	}
+
+	return newServiceEnvironment, nil
+}
+
+func WatchInstancers(l log.Logger, co Options, e Environment) {
```
kristinaspring, line 227: Does this need to be a public function? Presumably those using this package are going to call `NewEnvironment()`, right?

service/consul/environment.go
```diff
@@ -209,13 +208,69 @@ func NewEnvironment(l log.Logger, registrationScheme string, co Options, eo ...s
...
 				eo,
 				service.WithRegistrars(r),
 				service.WithInstancers(i),
 				service.WithCloser(closer),
-			)...,
-		), NewClient(consulClient)}, nil
+			)...), NewClient(consulClient)}
+
+	if co.DatacenterWatchInterval > 0 {
+		go WatchInstancers(l, co, newServiceEnvironment)
```
kristinaspring, line 221: Personally, I've been moving toward Start() functions to start any long running goroutines, and Stop() functions to be able to properly stop them.  This helps with following the general advice to know when a goroutine exits:
https://github.com/golang/go/wiki/CodeReviewComments#goroutine-lifetimes

I'm not sure if that's needed here or not.

service/environment.go
```diff
@@ -141,7 +145,35 @@ func (e *environment) DefaultScheme() string {
...
+	// add new instancers
+
+	for key, value := range instancersToAdd {
+		e.lock.Lock()
+		e.instancers.Set(key, value)
+		e.lock.Unlock()
+	}
+
+	// remove outdated instancers
+	for key := range e.instancers {
```
kristinaspring, line 164: shouldn't we be in at least a read lock in order to run through all the instancers? we're not looking at the values, but technically the keys could be changing, right?

kristinaspring (changes requested):

service/environment.go
johnabass: The set of instancers, along with most other Environment state, should be immutable after configuration is read.

service/consul/environment.go
johnabass: I agree.  That's a nice pattern that lots of other libraries follow.  This code was written before a lot of those de facto standards became a thing.

It would be nice to backport this code to use the Start/Stop idiom, but only if the level of effort isn't too bad.

service/environment.go
johnabass: Actually now I see what's going on with `UpdateInstancers`.  Yeah, you're right, a `sync.RWMutex` should probably be used here to guard the set of instancers.

service/consul/environment.go
johnabass: The idea behind exposing some of these other functions was that clients might want to reuse pieces of the package.  That wound up not happening.

This is probably a bigger issue.  This area of `webpa-common` really needs some refactoring to remove `zookeeper` support and simplify the watcher code.

service/consul/datacenterWatch.go
```diff
@@ -0,0 +1,109 @@
...
+	shutdown      chan struct{}
+	environment   Environment
+	options       Options
+}
+
+var (
+	defaultLogger = log.NewNopLogger()
+)
+
+func newDatacenterWatcher(logger log.Logger, environment Environment, options Options) (*DatacenterWatcher, error) {
```
kristinaspring, line 27: Is `newDatacenterWatcher()` meant to be unexported? No other package needs it?

service/consul/datacenterWatch.go
```diff
@@ -0,0 +1,109 @@
...
+			return
+		case <-checkDatacenters.C:
+			currentInstancers := environment.Instancers()
+			keys := make(map[string]bool)
+			instancersToAdd := make(service.Instancers)
+
+			datacenters, err := getDatacenters(logger, client, options)
+
+			if err != nil {
+				continue
```
kristinaspring, line 75: Do we need to log an error here? what happens if we always fail to get datacenters?

service/consul/datacenterWatch.go
j-mai: Hmmm yeah I left this as unexported because I envisioned that it was only going to be used when one creates a new consul service discovery environment. But I guess one could create a service discovery environment and then later create a datacenterWatcher without creating it in the `NewEnvironment` function. Although this would require whoever is using this datacenterWatcher to change the `DatacenterWatchInterval` duration from 0 to a number greater than 0 after it is passed into `NewEnvironment`, and I'm not sure why anybody would do that

service/consul/datacenterWatch.go
j-mai: Ah so my reasoning for adding a log here because if the `getDatacenters` errors, it already logs that inside that function, so I thought an additional log wasn't needed here. Would it be better to add a log here so that we know exactly what is calling `getDatacenters` if it errors?

kristinaspring (approved):

johnabass (approved):
This looks good!  there's a few opportunites for performance tweaks, but I don't know that those are important right now.  In fact, this may be just fine without it.

PR #538, opened 2020-11-05
added capabilities map and access to parsed values

kristinaspring:
Added `CapabilitiesMap`, a new implementation of the `CapabilitiesChecker` interface that the `MetricValidator` wraps (The `MetricValidator` provides a [bascule.ValidatorFunc](https://github.com/xmidt-org/bascule/blob/main/validator.go#L35).).  This allows us to have our capability metrics and only run regex against the request URL once, with the `MetricValidator` passing some values it parses to `CapabilitiesChecker`s.

The `CapabilitiesMap` uses the parsed endpoint value to determine which `CapabilityChecker` to run against the capabilities in the auth provided.  If there is no `CapabilityChecker` for the endpoint, the default is used.  This allows for us to have specific `CapabilityChecker`s for each endpoint a service supports, while having the same middleware handle all endpoints.


basculechecks/metricvalidator.go
```diff
@@ -32,7 +32,12 @@ var defaultLogger = log.NewNopLogger()
 // authorized given a bascule.Authentication object.  If it's not authorized,
 // a reason and error are given for logging and metrics.
 type CapabilitiesChecker interface {
-	Check(auth bascule.Authentication) (string, error)
+	Check(auth bascule.Authentication, vals ParsedValues) (string, error)
+}
+
+type ParsedValues struct {
+	Endpoint string
```
joe94, line 39: It might be worth adding a comment to explain the meaning we it give and how it's generated. Something like:
```
// Endpoint represents a bucket of URLs to be validated. An endpoint could be the regex string (i.e. `.*/api/v1/account/.*`) that matched the URL of a particular HTTP request whose capabilities need to be validated. 
```
Same with partner

kcajmagic (approved):

joe94 (approved):

basculechecks/capabilitiesmap.go
```diff
@@ -0,0 +1,99 @@
...
+// be some regex values, allowing for bucketing of urls that contain some kind
+// of ID or otherwise variable portion of a URL.
+type CapabilitiesMap struct {
+	checkers       map[string]CapabilityChecker
+	defaultChecker CapabilityChecker
+}
+
+// NewCapabilitiesMap validates the default Checker and sets up the map to be
+// non-nil in order to create a good CapabilitiesMap struct.
+func NewCapabilitiesMap(checkers map[string]CapabilityChecker, defaultChecker CapabilityChecker) (*CapabilitiesMap, error) {
```
johnabass, line 43: You're probably fine without this constructor.  You can make `checkers` and `defaultChecker` exported, and just let clients build a `CapabilitiesMap` directly.

basculechecks/capabilitiesmap.go
```diff
@@ -0,0 +1,99 @@
...
+	if defaultChecker == nil {
```
johnabass, line 44: Don't forget that a `nil` map or slice is valid for reading, just not for writing.  You can allow `nil` to let your client not bother creating things s/he doesn't need.

https://play.golang.org/p/fZ4Oo_ofkaE

basculechecks/capabilitiesmap_test.go
```diff
@@ -0,0 +1,210 @@
...
+	defaultCapabilities := []string{
+		"test",
+		"",
+		"default checker",
+		"...",
+	}
+	defaultToken := bascule.NewToken("test", "princ",
+		bascule.NewAttributes(map[string]interface{}{CapabilityKey: defaultCapabilities}))
+	badToken := bascule.NewToken("", "", nil)
+	tests := []struct {
```
johnabass, line 119: excellent use of table-driven tests
👍

johnabass (changes requested):

kcajmagic (approved):

basculechecks/metricvalidator.go
```diff
@@ -29,10 +29,26 @@ import (
...
+type ParsedValues struct {
+	// Endpoint is the string representation of a regular expression that
+	// matches the URL for the request.  The main benefit of this string is it
+	// most likely won't include strings that change from one request to the
+	// next (ie, device ID).
+	Endpoint string
+	// Partner is a string representation of the list of partners found in the
+	// JWT, where:
+	//   - any list including "*" as a partner is determined to be "wildcard".
+	//   - when the list is <1 item, the partner is deteremined to be "none".
```
joe94, line 48:
```suggestion
	//   - when the list is <1 item, the partner is determined to be "none".
```

joe94 (approved):
Looks good! Only a typo fix

johnabass (approved):
much cleaner!

PR #543, opened 2020-11-09
Fanout control

j-mai:
This PR allows for the ability to pull inactive datacenters from argus and update the Instancers based off the inactive datacenters saved in the database.

Closes xmidt-org/scytale#113


service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
-	environment := i.environment
-	client := i.environment.Client()
-	logger := i.logger
-	options := i.options
+func (d *DatacenterWatcher) watchDatacenters(ticker *time.Ticker) {
 
-	checkDatacenters := time.NewTicker(i.watchInterval)
+	client := d.environment.Client()
+	logger := d.logger
+	options := d.options
```
kcajmagic, line 124:
```suggestion
```
I don't think we should create these. It doesn't gain us much and can make it harder to update the code later.

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
+func (d *DatacenterWatcher) StartConsulTicker() {
+	if d.consulDatacenterWatch != nil {
+		ticker := time.NewTicker(d.consulDatacenterWatch.watchInterval)
+		go d.watchDatacenters(ticker)
+	}
 }
 
-func (i *datacenterWatcher) stop() {
-	close(i.shutdown)
+func (d *DatacenterWatcher) StopConsulTicker() {
```
kcajmagic, line 102: There should be one stop function that does all the cleanup.

kcajmagic: in this case we should leverage `func WithCloser(f func() error) Option `

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
+	if d.consulDatacenterWatch != nil {
+		close(d.consulDatacenterWatch.shutdown)
+	}
 }
 
-func (i *datacenterWatcher) watchDatacenters() {
+func (d *DatacenterWatcher) StartChrysomTicker() {
```
kcajmagic, line 108: After thinking about the stop functionality. Once the chrysom ticker is created it should start.

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
+	"github.com/xmidt-org/argus/model"
 	"github.com/xmidt-org/webpa-common/logging"
 	"github.com/xmidt-org/webpa-common/service"
 )
 
-//datacenterWatcher checks if datacenters have been updated, based on an interval
-//If they have, then the datacenterWatcher will update the Instancers dictionary
-type datacenterWatcher struct {
+//DatacenterWatcher checks if datacenters have been updated, based on an interval
+type DatacenterWatcher struct {
```
kcajmagic, line 18: This struct can be private
```suggestion
type datacenterWatcher struct {
```

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
 			}
+		}
+	}
+
+	environment.UpdateInstancers(keys, instancersToAdd)
+
+}
 
-			environment.UpdateInstancers(keys, instancersToAdd)
+func (d *DatacenterWatcher) DatacentersListener() chrysom.ListenerFunc {
```
kcajmagic, line 193:
```suggestion
func (d *datacenterWatcher) datacentersListener() chrysom.ListenerFunc {
```
This can be private and set directly to the chrysom config

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
+	return func(items []model.Item) {
+		updateInactiveDatacenters(items, d.inactiveDatacenters, &d.lock)
 
+		datacenters, err := getDatacenters(d.logger, d.environment.Client(), d.options)
+
+		if err == nil {
+			d.UpdateInstancers(datacenters)
```
kcajmagic, line 200:
```suggestion
```
chrysom and consul should not be coupled.

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
 		}
 
 	}
 }
+
+func updateInactiveDatacenters(items []model.Item, inactiveDatacenters map[string]bool, lock *sync.RWMutex) {
+	chrysomMap := make(map[string]bool)
+	for _, item := range items {
+		datacenterName := item.Data["name"].(string)
```
kcajmagic, line 209: a struct should be created so we can leverage json.Marshal to populate the struct. 
Such as:
```golang
type DatacenterFilter struct {
    Name string
    Inactive bool
}
```

This allows for flexibility for updating the filtering for later.

kcajmagic (changes requested):

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,229 @@
...
+		if ctx == nil {
+			ctx = context.Background()
+		}
+
+		datacenterWatcher.chrysomDatacenterWatch = &chrysomDatacenterWatch{
+			chrysomClient: chrysomClient,
+			ctx:           ctx,
+		}
+	} else if options.ChrysomConfig != nil {
+		return nil, errors.New("chrysom pull interval cannot be 0")
```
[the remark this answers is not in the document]
j-mai: This checked that if a chrysom config is declared but the pull interval is 0, then it would throw an error. Hopefully with the new changes this is clearer!

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,200 @@
...
+
+				//check if datacenter is part of inactive datacenters list
+				d.lock.RLock()
+				_, found := d.inactiveDatacenters[datacenter]
+				d.lock.RUnlock()
+
+				if found {
+					d.logger.Log(level.Key(), level.InfoValue(), logging.MessageKey(), "datacenter set as inactive", "datacenter name: ", datacenter)
+					continue
 				}
```
kcajmagic, line 139:
```suggestion
				d.lock.RLock()
				if _, found := d.inactiveDatacenters[datacenter]; found {
					d.logger.Log(level.Key(), level.InfoValue(), logging.MessageKey(), "datacenter set as inactive", "datacenter name: ", datacenter)
					continue
				}
				d.lock.RUnlock()
```

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,199 @@
...
-}
+	datacenterWatcher := &datacenterWatcher{
+		consulWatchInterval: options.DatacenterWatchInterval,
+		logger:              logger,
+		options:             options,
+		environment:         environment,
+		inactiveDatacenters: make(map[string]bool),
+	}
 
-func (i *datacenterWatcher) start() {
```
kristinaspring, line 46: Why get rid of the start function?

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,199 @@
...
+		if err != nil {
+			return nil, err
+		}
 
-	checkDatacenters := time.NewTicker(i.watchInterval)
+		//create chrysom client and start it
+		datacenterWatcher.chrysomClient = chrysomClient
+		datacenterWatcher.chrysomClient.Start(context.Background())
 
+	} else if options.ChrysomConfig != nil && options.ChrysomConfig.PullInterval <= 0 {
```
kristinaspring, line 82: I think moving this case earlier will make the code easier to understand.

Instead of having:
```
if config != nil && pullInterval > 0 {
...
} else if config != nil && pullInterval <= 0 {
return
}
...
```

We can have something like this:
```
if config != nil {
    if pullInterval <= 0 {
        return 
    }
    ...
}
...
```

It would fit in with the nil check for the `environment.Provider()` as well.

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,199 @@
...
 
-func (i *datacenterWatcher) start() {
-	go i.watchDatacenters()
-}
+	if options.ChrysomConfig != nil && options.ChrysomConfig.PullInterval > 0 {
 
-func (i *datacenterWatcher) stop() {
-	close(i.shutdown)
-}
+		if environment.Provider() == nil {
```
kristinaspring, line 59: We only use this `Provider()` when we use chrysom?

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,199 @@
...
+	if d.chrysomClient != nil {
+		d.chrysomClient.Stop(context.Background())
+	}
+}
+
+func (d *datacenterWatcher) watchDatacenters(ticker *time.Ticker) {
 	for {
 		select {
-		case <-i.shutdown:
+		case <-d.environment.Closed():
```
kristinaspring, line 103: It seems a little weird to use another object's shutdown channel? What's the benefit of doing this?

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,199 @@
...
-		case <-checkDatacenters.C:
-			currentInstancers := environment.Instancers()
-			keys := make(map[string]bool)
-			instancersToAdd := make(service.Instancers)
-
-			datacenters, err := getDatacenters(logger, client, options)
+		case <-ticker.C:
+			datacenters, err := getDatacenters(d.logger, d.environment.Client(), d.options)
 
 			if err != nil {
```
kristinaspring, line 110: What happens if we always get an error? Is there any way of knowing that datacenters aren't being updated?

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,199 @@
...
+
+	for _, w := range d.options.watches() {
+		if w.CrossDatacenter {
+			for _, datacenter := range datacenters {
+
+				//check if datacenter is part of inactive datacenters list
+				d.lock.RLock()
+				if _, found := d.inactiveDatacenters[datacenter]; found {
+					d.logger.Log(level.Key(), level.InfoValue(), logging.MessageKey(), "datacenter set as inactive", "datacenter name: ", datacenter)
+					continue
```
kristinaspring, line 135: `continue` is called before `RUnlock()`... I don't think this will work?

kristinaspring: My suggestion is to have one `RUnlock()` and separate the if:
```
d.lock.RLock()
_, found := d.inactiveDatacenters[datacenter]
d.lock.RUnlock()
if found {
    log
    continue
}
```

service/consul/datacenterWatch.go
```diff
@@ -1,109 +1,199 @@
...
+
+				w.QueryOptions.Datacenter = datacenter
+
+				// create keys for all datacenters + watched services
+				key := newInstancerKey(w)
+				keys[key] = true
+
+				// don't create new instancer if it is already saved in environment's instancers
+				if currentInstancers.Has(key) {
+					continue
```
kristinaspring, line 147: All these `continue`s give me the feeling that this code in the `for` loop would be better as its own function.  Not sure how easy that would be with various variables, but it might be worth a try! I think it might make the code easier to understand.

service/consul/options.go
```diff
@@ -18,6 +19,7 @@ type Watch struct {
 
 type Options struct {
 	Client                  *api.Config                    `json:"client"`
+	ChrysomConfig           *chrysom.ClientConfig          `json:"chrysomConfig"`
```
kristinaspring, line 22: Why does this need to be a pointer?

service/consul/datacenterWatch.go
kristinaspring: update: I would suggest adding a comment here explaining that `getDatacenters()` logs errors itself.  I do think we should add a metric here that tells us either 1) that we got datacenters or 2) the last time we successfully got them.  Otherwise, we won't have any visibility that this isn't working until something else goes wrong.  Adding this metric could be a different pull request.

kristinaspring: We had a weird issue in caduceus because of some code very similar to this.

service/consul/datacenterWatch.go
kristinaspring: Read further and the answer is yes...I think we should move this, but only because I think we should add a metric to this struct (see another comment below).  Could add a comment for now, explaining that only the chrysom client has metrics.

kristinaspring (changes requested):

service/consul/datacenterWatch.go
kcajmagic: Since the datacenterWatcher is part of the consul environment. When the Consul environment is closed, all of it supporting go routines should stop. This channel is so already available for all go routines to watch if need be.

service/consul/datacenterWatch.go
kcajmagic: I believed this should be removed, just as the consul client starts when created, the datacenterWatcher should start at the same time.

kcajmagic (approved):

kristinaspring (approved):
looks great! 🙂

johnabass (approved):
looks good!

PR #544, opened 2020-11-09
Fix bug in which tls could not be enabled

joe94:
Found this while attempting to run the webhook-transition code and realizing it always fails when only TLS is requested.

This is related to https://github.com/xmidt-org/webpa-common/pull/529https://github.com/xmidt-org/webpa-common/pull/529. I think there might've been some confusion about the `generateCerts` function name. I renamed it to `loadCerts` as all it does is validate cert/key pairs and loads them (not generate them).


j-mai (approved):
looks good to me!

kcajmagic (approved):

PR #547, opened 2020-11-11
Gate Filtering

j-mai:
This PR:

- Allows for filters to be added to keep devices that match certain filters from connecting
- Handler to get and edit saved filters


device/devicegate/filter.go
```diff
@@ -0,0 +1,263 @@
...
+	allowedFilters map[string]struct{}
+	lock           sync.RWMutex
+}
+
+type filterRequest struct {
+	Key    string
+	Values []interface{}
+}
+
+func NewFilterGate(options ...FilterGateOption) *FilterGate {
```
johnabass, line 38: You can dramatically simplify this code if you avoid the use of a constructor and supply and interface that clients uses the `FilterGate` as.  That separates creation time from run time.  For example:

```
type Gate interface {
    // whatever methods you want to expose to clients here:
    Filters() map[string]interface{}
}

type FilterGate struct {
    // export things to simplify construction
    Filters map[string]map[interface{}]struct{}

    //etc
}

// make *FilterGate implement Gate ...

// clients can construct a FilterGate directly, but use it as a Gate:
func NewGate() Gate {
    return &FilterGate{
        Filter: ...
    }
}
```

device/devicegate/filter.go
```diff
@@ -0,0 +1,263 @@
...
+)
+
+var (
+	emptyVal = struct{}{}
+)
+
+type FilterGateOption func(*FilterGate)
+
+type FilterGate struct {
+	filters        map[string]map[interface{}]struct{}
```
johnabass, line 28: This is complex enough to warrant a custom type.  Also, what purpose is the `map[interface{}]struct{}`?  If you're wanting a set of `interface{}`, a better type declaration is `map[interface{}]bool`.

device/devicegate/filter.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+
+		copy[k] = filters
+	}
+	f.lock.RUnlock()
+
+	return copy
+}
+
+func (f *FilterGate) FiltersCopy() map[string]map[interface{}]struct{} {
```
johnabass, line 79: A better design is to make is so that you never need to hand out the filters data structure, and thus never need to copy it.  For example, you could have methods like so:

```
// (consider simplifying so you don't need a map of maps)
func (f *FilterGate) Filters(key string) (map[interface{}]struct{}, bool) {
    f.lock.RLock()
    defer f.lock.RUnlock()
    v, ok := f.filters[key]
    return v, ok
}
```

device/devicegate/filter.go
```diff
@@ -0,0 +1,263 @@
...
+	return filtersCopy
+}
+
+func (f *FilterGate) Filters() map[string]map[interface{}]struct{} {
+	return f.filters
+}
+
+//allows for adding and removing filters
+//Will completely overwrite current filters in place
+func (f *FilterGate) EditFilters(key string, values []interface{}, add bool) bool {
```
johnabass, line 103: See my earlier comment about not exposing the filters map.  You can supply an API that's easier to use with methods like:

```
Set(key string, value ...interface{}) []interface{} // return the old values, you may also want to return a bool
Delete(key string)
```

device/devicegate/filter.go
```diff
@@ -0,0 +1,263 @@
...
+			}
+		}
+
+	}
+
+	return true, device.MatchResult{}
+}
+
+func metadataMapMatch(keyToCheck string, filterValues map[interface{}]struct{}, m *device.Metadata) bool {
+	metadataVal := m.Load(keyToCheck)
```
johnabass, line 157: If you make a custom type for the filters, something like `type Filters map[string]map[interface{}]struct{}`, then functions like this can simply be methods of that type.

device/devicegate/filter.go
```diff
@@ -0,0 +1,263 @@
...
+		if found {
+			return true
+		}
+	}
+
+	return false
+
+}
+
+func (f *FilterGate) ServeHTTP(response http.ResponseWriter, request *http.Request) {
```
johnabass, line 202: A better approach is to use black box reuse or composition.  Implement an `http.Handler` that contains a `*FilterGate`:

```
type FilterHandler struct {
    Gate Gate // use an interface is preferred
}

func (fh FilterHandler) ServeHTTP(...) {
    ...
}
```

That way is easier to unit test and maintain over time, since adding features to the `Gate` won't necessarily affect the `http.Handler` code.

device/manager.go
```diff
@@ -86,9 +104,12 @@ type Manager interface {
 	Registry
 }
 
+// ManagerOption is a configuration option for a manager
+type ManagerOption func(*manager)
+
 // NewManager constructs a Manager from a set of options.  A ConnectionFactory will be
 // created from the options if one is not supplied.
-func NewManager(o *Options) Manager {
+func NewManager(o *Options, options ...ManagerOption) Manager {
```
johnabass, line 112: A simpler approach is to make `Filter` a field of the `Options` struct.  Then, move the creation semantics for the `Filter` somewhere else.  That keeps the manager code simple and easier to maintain.

johnabass (changes requested):
I have a few more comments but this is good for a first round.

device/manager.go
```diff
@@ -181,6 +222,11 @@ func (m *manager) Connect(response http.ResponseWriter, request *http.Request, r
 		Logger:     m.logger,
 	})
 
+	if allow, matchResults := m.filters.AllowConnection(d); !allow {
+		d.infoLog.Log("filter", "filter match found,", "location", matchResults.Location, "key", matchResults.Key)
+		return nil, nil
```
j-mai, line 227: return an error when filter match is found

device/manager.go
```diff
@@ -127,7 +148,15 @@ func NewManager(o *Options) Manager {
 		listeners:             o.listeners(),
 		measures:              measures,
 		enforceWRPSourceCheck: wrpCheck.Type == CheckTypeEnforce,
+		filters:               defaultFilterFunc(),
```
j-mai, line 151: plural is confusing

device/devicegate/filter.go
```diff
@@ -0,0 +1,241 @@
...
+type Set interface {
+
+	// Has returns true if a value exists in the set, false if it doesn't.
+	Has(interface{}) bool
+
+	// VisitAll applies the visitor function to every value in the set.
+	VisitAll(func(interface{}))
+
+	// String returns a string representation of the set.
+	String() string
```
kcajmagic, line 53: Since the `FilterSet` is returning a json representation of the Set. Have this interface include [`json.Marshaler`](https://golang.org/pkg/encoding/json/#Marshaler)

device/devicegate/filterHandler.go
```diff
@@ -0,0 +1,123 @@
...
+		fmt.Fprintf(b, `"%s": `, key)
+		fmt.Fprintf(b, "%s", val.String())
+		needsComma = true
+
+		return true
+	}
+}
+
+// wrapper to build JSON string representation of filters
+func filtersToString(g Interface) string {
```
kcajmagic, line 110: I think it would cleaner to implement  [`json.Marshaler`](https://golang.org/pkg/encoding/json/#Marshaler)

kcajmagic (changes requested):

kcajmagic (approved):
Looks good

device/devicegate/filter.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (s FilterSet) VisitAll(f func(interface{})) {
+	for key := range s {
+		f(key)
+	}
+}
+
+func (s FilterSet) MarshalJSON() ([]byte, error) {
+	temp := make([]interface{}, len(s))
```
johnabass, line 163: VisitAll holds a lock, but this method does not.  That means that elements of the slice can be updated concurrently to this method.  Is that ok?  Consider this scenario:

```
var s FilterSet // initialize somehow
myValue := new(int)
s["key"] = myValue
go s.MarshalJSON()
go func() {
    *myValue = 123 // an update that may occur simultaneously with MarshalJSON
}()
```

I can't tell if that's an issue, however.  Do we anticipate modifying values *after* they are put into a `FilterSet`?  If so, then `MarshalJSON` and any other mutating method will need to hold a lock.

device/devicegate/filter.go
```diff
@@ -0,0 +1,212 @@
...
+	index := 0
+	s.VisitAll(func(v interface{}) {
+		temp[index] = v
```
johnabass, line 166: You can simplify this closure by removing the index counter.  A slice tuple can effectively be your index:

```
temp := make([]interface{}, 0, len(s)) // note the second parameter
s.VisitAll(func(v interface{}) {
    temp = append(temp, v)
})
```

device/devicegate/filterHandler.go
```diff
@@ -0,0 +1,88 @@
...
+// FilterHandler is an http.Handler that can get, add, and delete filters from a devicegate Interface
+type FilterHandler struct {
+	Gate Interface
+}
+
+func (fh *FilterHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
+	logger := logging.GetLogger(request.Context())
+
+	method := request.Method
+	if method == http.MethodGet {
```
johnabass, line 24: In general, you don't want to check for things like HTTP method to perform custom logic.  Things like reading the body and parsing the URL form variables works properly no matter what HTTP method is used.  Leverage HTTP routing code, such as `http.ServeMux` or `gorilla/mux.Router`, to make decisions about what type of handling should be done.  That makes your HTTP handling code very generic and composable.

Here's two approaches to this.  Both are great, and it's a matter of personal taste which you use.

Approach 1: Different HTTP handlers
```
type MyHandler struct { /* a ServeHTTP method expecting a GET */ }
type FormHandler struct { /* a ServeHTTP method that looks at the content body */ }

r := mux.NewRouter()
r.Path("/foo").Methods("GET").Handler(new(MyHandler))
r.Path("/foo").Methods("POST").Handler(new(FormHandler))
```

Approach 2: Methods of the same type as handlers
```
type MyHandler struct {
    // stuff ....
}

func (mh *MyHandler) Simple(response http.ResponseWriter, request *http.Request) {
    // snip
}

func (mh *MyHandler) Form(response http.ResponseWriter, request *http.Request) {
    // snip
}

h := new(MyHandler)
r := mux.NewRouter()
r.Path("/foo").Methods("GET").HandlerFunc(h.Simple)
r.Path("/foo").Methods("POST").HandleFunc(h.Form)
```

device/devicegate/filterHandler.go
```diff
@@ -0,0 +1,88 @@
...
+			} else {
+				response.WriteHeader(http.StatusOK)
+			}
+		} else if method == http.MethodDelete {
+			fh.Gate.DeleteFilter(message.Key)
+			response.WriteHeader(http.StatusOK)
+		}
+
+		filtersJSON, _ := json.Marshal(fh.Gate)
+		logger.Log(level.Key(), level.InfoValue(), logging.MessageKey(), "gate filters updated", "filters", string(filtersJSON))
```
johnabass, line 63: General purpose logging like this makes better sense as a decorator.  That makes this code less noisy and avoids repeating yourself.  For example:

```
type GateLogger struct {
    Logger log.Logger
}

func (gl GateLogger) Then(next http.handler) http.Handler {
    return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
        // first, delegate to the next handler to do whatever is necessary for the gate
        next.ServeHTTP(response, request)

        // now, perform the logging.  the "filtersJSON" can be communicated through the request's context
        // it's probably easiest to set the entire Gate object in the context inside the next handler or another decorator

        if filtersJSON, err := json.Marshal(/* get the Gate from request.Context() */); err == nil {
            logger.Log(level.Key(), level.InfoValue(), logging.MessageKey(), "gate filters updated", "filters", string(filtersJSON))
        } else {
            // NEVER SWALLOW ERRORS! log the error here
            // sometimes, it is appropriate to panic since this indicates that something's buggy with the gate
        }
    })
}

// You can use MiddlewareFunc to decorate things
gl := GateLogger{
    Logger: /* logger goes here! */
}

r := mux.Router()
r.MiddlewareFunc(gl.Then) // this will apply the middleware to any routes defined afterward

// You can also use alice to decorate each handler separately
c := alice.NewChain(gl.Then)
r.Path("/foo").Handle(c.Then(myHandler))
r.Path("/bar").Handle(nonGateHandler)
```

johnabass (changes requested):

device/devicegate/filter.go
j-mai: I intended for FilterSet to implement the Set interface, which is supposed to be a read-only Set data structure. So the intention was for FilterSet to be read-only. But if someone were to use FilterSet by itself and not as a Set interface, they certainly could modify values after they put it into the FilterSet, so it would probably be safer to include a lock here

johnabass (approved):
That looks like a great first version.

PR #549, opened 2020-12-05
avoid duplicate metric name

joe94:


joe94:
Spoke with @kcajmagic and concluded this will introduce a lot of unnecessary work for updating our different dashboards. Since we only needed this change because of Hecate, we'll go ahead and let Hecate use argus/chrysom directly instead of relying on xwebhook.

PR #556, opened 2020-12-22
add provideMetrics for AWS and webhooks

j-mai:


webhook/aws/sns.go
```diff
@@ -62,7 +61,7 @@ type SNSServer struct {
 // Notifier interface implements the various notification server functionalities
 // like Subscribe, Unsubscribe, Publish, NotificationHandler
 type Notifier interface {
-	Initialize(*mux.Router, *url.URL, string, http.Handler, log.Logger, xmetrics.Registry, func() time.Time)
+	Initialize(*mux.Router, *url.URL, string, http.Handler, log.Logger, AWSMetrics, func() time.Time)
```
kristinaspring, line 64: This might need to be a pointer? I'm not sure; if it works though, great!

kristinaspring (approved):

webhook/aws/sns.go
j-mai: The `SNSServer` struct didn't have AWSMetrics as a pointer so I decided to pass it in by value

PR #557, opened 2021-01-05
Update xwebhook to latest Argus codebase

joe94:
As part of the docker harmonize change, we want to have a docker compose cluster that works with latest Argus@main. This change updates the xwebhook logic that we can import into Tr1d1um and Caduceus.


joe94:
@j-mai I added you too as I had to update some of the datacenter code you had written.

kcajmagic (approved):

j-mai (approved):

PR #558, opened 2021-01-06
CounterVec version of bascule metric providers

joe94:
In order to partition metrics based on the different servers the bascule profiles (introduced in https://github.com/xmidt-org/argus/issues/97) target, we need the CounterVecs so we can leverage https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec.CurryWith

### DIFF
--- a/basculechecks/metrics.go
+++ b/basculechecks/metrics.go
@@ -63,7 +63,7 @@ const (
 	EmptyParsedURL           = "empty_parsed_URL"
 )
 
-// Metrics returns the Metrics relevant to this package targetting our older non uber/fx applications.
+// Metrics returns the Metrics relevant to this package targeting our older non uber/fx applications.
 // To initialize the metrics, use NewAuthCapabilityCheckMeasures().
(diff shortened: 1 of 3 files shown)


basculechecks/metrics.go
```diff
@@ -63,7 +64,7 @@ func Metrics() []xmetrics.Metric {
 			Name:       AuthCapabilityCheckOutcome,
 			Type:       xmetrics.CounterType,
 			Help:       "Counter for the capability checker, providing outcome information by client, partner, and endpoint",
-			LabelNames: []string{OutcomeLabel, ReasonLabel, ClientIDLabel, PartnerIDLabel, EndpointLabel},
+			LabelNames: []string{ServerLabel, OutcomeLabel, ReasonLabel, ClientIDLabel, PartnerIDLabel, EndpointLabel},
```
joe94, line 67: This will force existing consumers to add the value for the label. We could add it only for the CounterVec case if we wanted.

kristinaspring (approved):

PR #562, opened 2021-01-27
Standardize measures building a bit

joe94:
Creating measures for validation and capability the same way will make the code a bit more standardized/readable.


kristinaspring (approved):
🚀

PR #573, opened 2021-04-01
Changes for compatibility with Chrysom Otel changes.

utsavbatra5:


joe94:
We'll have to update the code in the `xwebhook` package as well. Passing in `nil` as the `getLogger` func and `context.Background()` should suffice. That code should be deprecated and we just want it to compile.

utsavbatra5:
Sure Joe. I have now added the above additional changes to this PR. Kindly review it.

joe94 (approved):

PR #579, opened 2021-05-11
Build metric listener directly

joe94:
While working on https://github.com/xmidt-org/argus/pull/174, I found that we don't need to build `AuthValidationMeasures` independently. All current client code only needs it to build the metric listener.


kristinapathak (approved):
👍

PR #582, opened 2021-07-06
Cycle

kristinapathak:


j-mai (approved):
looks good to me!

service/consul/datacenterWatch.go
```diff
@@ -65,16 +65,17 @@ func newDatacenterWatcher(logger log.Logger, environment Environment, options Op
...
 
 		options.ChrysomConfig.Listen.Listener = datacenterListenerFunc
-
 		options.ChrysomConfig.Logger = logger
-		chrysomClient, err := chrysom.NewClient(options.ChrysomConfig, nil)
+
+		m := &chrysom.Measures{
+			Polls: environment.Provider().NewCounterVec(chrysom.PollCounter),
+		}
+		chrysomClient, err := chrysom.NewClient(options.ChrysomConfig, m, nil, nil)
```
joe94, line 78: Do we want to provide the setLogger and getLogger args here so logging can happen for `GetItems`?

kristinapathak: I wasn't trying to change any functionality - it was passing `nil` before so I'm just continuing that. I can change it, but I don't know anything about testing this code...

joe94: @j-mai, thoughts on this? It might be helpful for debugging purposes but I'm ok leaving as is as well since it wasn't there already

j-mai: Yeah I think passing a `getLogger` function a good idea! When I first wrote this code, there wasn't the option to pass in logging functions when creating the chrysom client, which is why the functionality isn't there right now. A question: I understand that `getLogger` provides the chrysom client with a logger to use, but why would passing a `setLogger` function be useful in this case?

kristinapathak: `setLogger` is determining where to put the logger in the context - `getLogger` can only get the logger if they both are accessing the same spot.

joe94 (approved):

joe94 (approved):

PR #586, opened 2021-08-23
don't dispatch events on index underflow or on retryable errors

johnabass:


service/consul/instancer.go
```diff
@@ -136,29 +145,32 @@ func (i *instancer) getInstances(lastIndex uint64, stop <-chan struct{}) ([]stri
...
-		if meta == nil || meta.LastIndex < lastIndex {
-			lastIndex = 0
-		} else {
-			lastIndex = meta.LastIndex
+		queryOptions.WaitIndex = lastIndex
+		entries, meta, resp.err = i.client.Service(i.service, i.tag, i.passingOnly, &queryOptions)
+		if resp.err == nil {
+			// see: https://www.consul.io/api-docs/features/blocking#implementation-details
+			if meta == nil || meta.LastIndex < lastIndex {
+				resp.err = errIndexUnderflow
```
schmidtw, line 160: Would there be any value to having metrics showing when there are boundary conditions like this occuring?

joe94, line 160: Based on the docs, it looks like we should be resetting the index to zero here:
```
resp.index = 0
```

"Implementations must check to see if a returned index is lower than the previous value, and if it is, should reset index to 0 - effectively restarting their blocking loop."

service/consul/instancer.go
```diff
@@ -136,29 +145,32 @@ func (i *instancer) getInstances(lastIndex uint64, stop <-chan struct{}) ([]stri
...
+			} else if meta.LastIndex == 0 {
+				resp.err = errIndexZero
```
joe94, line 162: and setting the index to 1 in this case:
```
resp.index = 1
```
"After the initial request (or a reset as above) the X-Consul-Index returned should always be greater than zero. [...] It is always safe to use an index of 1 to wait for updates when the data being requested doesn't exist yet, so clients should sanity check that their index is at least 1 after each blocking response is handled to be sure they actually block on the next request."

service/consul/instancer.go
```diff
@@ -136,29 +145,32 @@ func (i *instancer) getInstances(lastIndex uint64, stop <-chan struct{}) ([]stri
...
-		}
-
-		// see: https://www.consul.io/api-docs/features/blocking#implementation-details
-		if meta == nil || meta.LastIndex < lastIndex {
-			lastIndex = 0
-		} else {
-			lastIndex = meta.LastIndex
+		queryOptions.WaitIndex = lastIndex
+		entries, meta, resp.err = i.client.Service(i.service, i.tag, i.passingOnly, &queryOptions)
+		if resp.err == nil {
```
joe94, line 157: When `resp.err != nil`, does meta contain any info? Currently, we just reset the index back to in this case

joe94 (changes requested):
This is looking promising

service/consul/instancer.go
joe94: Nevermind, 0 is the default value.

schmidtw (approved):

joe94 (approved):

kcajmagic (approved):

service/consul/instancer.go
johnabass: Possibly.  Remember, though, that this is normal consul behavior.  In particular, this happens in the face of network partitioning, which is simply a temporary condition that service discovery infrastructure should be tolerant of.

If we want metrics around this, let's do it in a separate PR.

PR #588, opened 2021-10-28
Http error retry funcs

mtrinh11:


xhttp/mocks_test.go
```diff
@@ -56,3 +56,11 @@ func (m *mockListener) Close() error {
 func (m *mockListener) Addr() net.Addr {
 	return m.Called().Get(0).(net.Addr)
 }
+
+type mockTempError string
+
+func (m mockTempError) Temporary() bool { return true }
+
+func (m mockTempError) Timeout() bool { return true }
```
kristinapathak, line 64: Do we need this function?

kristinapathak (approved):
Looks good! Two small suggestions.  Otherwise, we can merge once we're sure the community is good with this change. 👍

xhttp/mocks_test.go
mtrinh11: I included Timeout so that mockTempError would implement the Error interface

kristinapathak (approved):

kristinapathak (approved):

PR #600, opened 2022-05-19
add utf-8 check for device handler

kristinapathak:


denopink (approved):
besides not being able to be tested (write/read pump) atm, lgtm 🍻 
We're already talking about refactoring the write/read pump for improve testability

Issue #655, opened 2022-09-16
Archival & Deprecation Roadmap


denopink:
This ticket will outline a general roadmap for archiving & deprecating webpa-common. Xmidt libraries depending on webpa-common will tag this ticket.

## Packages ready to be replacedackages that
- https://github.com/uber-go/fx will replace
  - secure*
  - service*
  - 

## Packages that need to be moved
- `device*` to https://github.com/xmidt-org/talaria/
- `bascule*` to https://github.com/xmidt-org/bascule

## Packages unaccounted for
**i.e.: which libs depends on these packages and what are we replacing them if they're still in use**
- bookkeeping
- capacitor
- clock
- concurrent
- convey*
- event
- hash
- health
- logging*
- middleware
- resource
- semaphore
- server
- tracing*
- transport*
- webhook*
- xhttp*
- xlistener*
- xmetrics
- xresolver*
- xviper*

denopink:
Addressing the deprecation of `github.com/xmidt-org/webpa-common/v2/logging/*` in affected servers/libs, updating ticket body with info

denopink:
as mentioned above, we're going forward with replacing both webpa-common/logging and go-kit/logging with zap

denopink:
reverting service/consul & service/servicecfg packages back to using gokit log because of gokitzk.NewRegistrar

PR #874, opened 2023-05-03
Zap logger updates

maurafortino:
What's included:

- https://github.com/xmidt-org/webpa-common/issues/655
  ** this won't deprecate webpa-common fully from scytale - but will remove the logging and go-kit features from scytale
- added zap logger to servicecfg.NewEnvironment to be used in Scytale
- added adapter struct to act as an interface so zap logger could be used in certain functions that required the Log interface


service/consul/registrar.go
```diff
@@ -69,23 +69,25 @@ func (tc ttlCheck) updatePeriodically(updater ttlUpdater, shutdown <-chan struct
...
 				successiveErrorCount++
 				if successiveErrorCount == 1 {
-					tc.logger.Log(level.Key(), level.ErrorValue(), "msg", "error while updating TTL to passing", "error", err)
+					e := zap.Error(err)
+					tc.logger.Error("error while updating TTL to passing", e)
 				}
 			} else if successiveErrorCount > 0 {
-				tc.logger.Log(level.Key(), level.InfoValue(), "msg", "update TTL success", "previousErrorCount", successiveErrorCount)
+				i := zap.Int("previousErrorCount", successiveErrorCount)
+				tc.logger.Info("update TTL success", i)
```
denopink, line 77:
```suggestion
				tc.logger.Info("update TTL success", zap.Int("previousErrorCount", successiveErrorCount))
```

service/consul/registrar.go
```diff
@@ -100,17 +102,17 @@ func appendTTLCheck(logger log.Logger, serviceID string, agentCheck *api.AgentSe
...
 			interval: interval,
-			logger: log.With(
-				logger,
-				"serviceID", serviceID,
-				"checkID", agentCheck.CheckID,
-				"ttl", agentCheck.TTL,
-				"interval", interval.String(),
+			logger: logger.With(
+				s, c, t, i,
 			),
```
denopink, line 116:
```suggestion
			logger: logger.With(
				zap.String("serviceID", serviceID), zap.String("checkID", agentCheck.CheckID), zap.String("ttl", agentCheck.TTL), zap.String("interval", interval.String()),
			),
```

service/consul/registrar.go
```diff
@@ -100,17 +102,17 @@ func appendTTLCheck(logger log.Logger, serviceID string, agentCheck *api.AgentSe
 		return nil, fmt.Errorf("TTL %s is too small", agentCheck.TTL)
 	}
 
+	s := zap.String("serviceID", serviceID)
+	c := zap.String("checkID", agentCheck.CheckID)
+	t := zap.String("ttl", agentCheck.TTL)
+	i := zap.String("interval", interval.String())
```
denopink, line 108:
```suggestion
```

service/consul/registrar_test.go
```diff
@@ -2,17 +2,21 @@ package consul
...
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
+	"github.com/xmidt-org/sallust"
+	"github.com/xmidt-org/webpa-common/v2/adapter"
 )
 
+var log = adapter.Adapter{
+	Logger: sallust.Default(),
+}
+
```
denopink, line 19: I think we want to avoid using global loggers like this

service/zk/environment.go
```diff
@@ -28,30 +26,33 @@ func newService(r Registration) (string, gokitzk.Service) {
 // clientFactory is the factory function used to create a go-kit zookeeper Client.
 // Tests can change this for mocked behavior.
 var clientFactory = gokitzk.NewClient
+var logger = adapter.Adapter{
+	Logger: sallust.Default(),
+}
```
denopink, line 31: I think we want to avoid global loggers like this

denopink (changes requested):
I left a few suggestions.
Also, there's a few formatting notes like newlines after `{}` blocks like `if` and `for`  and func blocks.
Looks like some webpa tests are failing

service/consul/environment.go
```diff
@@ -76,38 +68,39 @@ func newInstancerKey(w Watch) string {
...
 }
-
 func defaultClientFactory(client *api.Client) (Client, ttlUpdater) {
 	return NewClient(client), client.Agent()
 }
 
 var clientFactory = defaultClientFactory
+var logger = adapter.Adapter{
+	Logger: sallust.Default(),
+}
```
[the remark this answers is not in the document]
johnabass: A better approach than a global default instance would be to have a global `DefaultLogger` function that returns the adapter.

```
func DefaultLogger() *zap.Logger {
    return adapter.Adapter{
        Logger: sallust.Default(),
    }
}
```

This avoids having global mutable state.  It also leverages the fact that `golang` inlines short functions like this, so in the compiled binary this kind of function will just get compiled away in favor of local code.

adapter/adapter.go
```diff
@@ -0,0 +1,21 @@
+package adapter
+
+import "go.uber.org/zap"
+
+type Adapter struct {
```
johnabass, line 5: The type `adapter.Adapter` stutters.  It would be better to change this package name to something like `logger`, `logging`, `log`, etc.  There may also be an existing `webpa-common` package that would be appropriate.

service/consul/datacenterWatch.go
```diff
@@ -203,7 +203,11 @@ func createNewInstancer(keys map[string]bool, instancersToAdd service.Instancers
 
 	// don't create new instancer if it was already created and added to the new instancers map
 	if instancersToAdd.Has(key) {
-		dw.logger.Log(level.Key(), level.WarnValue(), "msg", "skipping duplicate watch", "service", w.Service, "tags", w.Tags, "passingOnly", w.PassingOnly, "datacenter", w.QueryOptions.Datacenter)
+		s := zap.String("service", w.Service)
+		t := zap.Any("tags", w.Tags)
+		p := zap.Bool("passingOnly", w.PassingOnly)
+		d := zap.String("datacenter", w.QueryOptions.Datacenter)
+		dw.logger.Warn("skipping duplicate watch", s, t, p, d)
```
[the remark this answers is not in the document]
johnabass: Owen's suggestion is a good one.

Something to consider, if possible:  you can add fields to a logger that are output with every line.  This works well when you have "constant" values that don't change but that you still want in each log line.  `datacenter` is a good candidate for that:

```
var a adapter.Adapter // assume this came from somewhere else

localAdapter := adapter.Adapter{
    Logger: a.Logger.With(zap.String("datacenter", datacenter)),
}
```

Then you can just store `localAdapter` as a struct field and use that logger/adapter for logging.

johnabass (changes requested):

johnabass (approved):

denopink (approved):
lgtm 🍻

PR #910, opened 2023-06-08
removed log.Valuer and updated each log individually to update eventCount

maurafortino:
currently receiving the following error in some monitor.go logs: "json: unsupported type: log.Valuer"
this fix will update the statement so that we see "eventCount: [number]"


denopink (approved):
lgtm! 🍻

PR #1101, opened 2025-02-17
patch: service hashing should use the host component from the registered consul url

denopink:
- previously we used the full registered consul url for hashing, leading to issues when the urls resulted in different hash rings across services
- our hashing should only consider the host component of the url


johnabass (approved):
Looks good.  Great find!

schmidtw (approved):

PR #1142, opened 2025-08-06
made reuse file cover the entire repo

maurafortino:


semaphore/doc.go
```diff
@@ -3,5 +3,9 @@
 
 /*
 Package semaphore provides a simple channel-based semaphore that optionally honors context semantics.
+
```
maurafortino, line 6: spoke with @johnabass about deprecating this package.

PR #1143, opened 2025-08-06
adding a LICENSE file to see if that fixes missing license issue

maurafortino:


maurafortino:
doing this based off the fact that wrp-go/v3 has a LICENSE file and does not have the same issues as webpa-common/v2 has. not sure if this will work. https://pkg.go.dev/github.com/xmidt-org/wrp-go/v3@v3.7.0

PR #1160, opened 2025-11-03
Add Intermediate-Context header support for Talaria message metadata

ChaitanyaSingla:
This PR introduces the `intermediate-context` header as an optional field in the metadata preparation flow for both online and offline messages sent by Talaria.

Key Changes: 
1. Added support for `intermediate-context` header in message metadata construction.
2. Updated relevant structs and interfaces to include the new header.

Purpose:
The `intermediate-context` header allows for carrying any extra information that may be required by the receiving systems.


Sachin4403 (approved):

schmidtw (approved):

denopink (approved):
I made 1 change, updated the implementation to match `device.newDevice`current options pattern for consistency

denopink:
fyi releasing and then updating https://github.com/xmidt-org/talaria/pull/558, if anything break then I'll patch it

